PR #21, opened 2020-01-23
Limit for timer event

manuel-fischer:
When a timer event is already enqueued in the event queue, another timer event should not be enqueued again.

### DIFF
--- a/source/tscript.js
+++ b/source/tscript.js
@@ -6817,8 +6817,7 @@ module.Interpreter = function(program)
 			}
 			if (this.background && !this.timerEventEnqueued)
 			{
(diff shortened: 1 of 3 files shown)


source/tscript.js
```diff
@@ -7072,11 +7078,16 @@ module.Interpreter = function(program)
...
+	// event limits
+	this.timerEventEnqueued = false;
 
 	// queue an event
+	// returns true when event got enqueued, false otherwise
 	this.enqueueEvent = function(type, event)
 	{
 		if (this.eventmode) this.eventqueue.push({"type": type, "event": event});
+		return this.eventmode;
 	};
```
manuel-fischer: Little change to the `enqueueEvent` function, it now returns true if the event got enqueued, this currently only happens when `this.eventmode` is true.

PR #26, opened 2020-02-20
Changed icons on toolbar

manuel-fischer:
More commonly used icons for new, open and save. It makes them more distinguishable. I often mixed them up, when trying to save the code to the local storage. Also tweaked the breakpoint icon to look a little bit better.

change


TGlas:
Thanks a lot, that looks really good!

PR #28, opened 2020-10-27
clarify open and save file dialogs

shakenes:
I tried to clarify the file dialogs because the windows are unintuitively empty if no documents are saved. In that case the user gets a note that there are no documents. Also, the filename input element gets a placeholder hinting the user that a new filename can be entered here. 

Disclaimer: I don't know JavaScript, so although it works, it might not be well done :)


TGlas:
Thanks, looks good!

PR #31, opened 2020-11-18
Add a delete button to file dialogs

Jartreg:
It's already possible to delete files by pressing the backspace or delete button in file dialogs. This adds an explicit button to the bottom. In save dialogs, where there is a text field for the file name, the field is moved above the buttons.

This was recently requested by someone in a Zoom call.

## Screenshots

Load file dialog

Save file dialog


TGlas:
Thanks, looks good!

PR #33, opened 2020-11-20
Add export import

Tilo-K:
This adds functionality to export/download a script as a ".tscript" file. As well as adding an Import to do the reverse.


source/ide.js
```diff
@@ -1660,6 +1675,42 @@ function fileDlg(title, filename, allowNewFilename, onOkay)
...
+		fileImport.addEventListener('change', async (event) => {
+			if(event.target.files){
+				var file = event.target.files[0];
+				var filename = file.name.split('.tscript')[0];
+				var data = await file.text();
+				localStorage.setItem("tscript.code." + filename, data);
+				files.push(filename);
+				var option = document.createElement('option');
+				option.text = filename;
+				list.add(option);
```
Jartreg, line 1708: [`HTMLSelectElement`](https://developer.mozilla.org/docs/Web/API/HTMLSelectElement) does't have an `add` method. You probably meant `list.options.add(option)`, but just changing this won't work either, because `list` might not be an `HTMLSelectElement` at all.

When there are no saved files, `list` is created as `<text>No documents saved.</text>` ([here](https://github.com/Tilo-K/tscript/blob/c1e0ab9692eb8d1928d56a143f5efad46d27a24c/source/ide.js#L1533)). This means that in this case, `list` is an [`HTMLUnknownElement`](https://developer.mozilla.org/docs/Web/API/HTMLUnknownElement), which obviously doesn't have an `options` parameter:

The resulting TypeError

source/index.html
```diff
@@ -41,6 +41,8 @@
 	<link rel="stylesheet" href="documentation.css">
 </head>
 <body>
+
 <div id="ide-container" style="position: fixed; left: 0; top: 0; width: 100vw; height: 100vh;"></div>
+</div>
```
Jartreg, line 46: The additional closing tag is unneeded. The div is already closed in the previous line.

```suggestion
<div id="ide-container" style="position: fixed; left: 0; top: 0; width: 100vw; height: 100vh;"></div>
```

Jartreg (commented):
Hi there! I have some questions/suggestions for your changes:

## Removed examples
Did you mean to include the distribution files and `source/def-examples.js`? By doing that you might've accidentally removed all examples from the documentation:

Missing examples error

## The file input element
In the beginning, you tried to add the `input` element to the page dynamically but then added it to the HTML markup instead. Now it's only in `distribution/end`, but not in `source/index.html`, which means that importing is broken in that file.

I think it would still be better if you added it dynamically. The issues you had probably came from the way you used [`document.createElement(...)`](https://developer.mozilla.org/docs/Web/API/Document/createElement) ([here](https://github.com/Tilo-K/tscript/blob/d9f0cdb84ddc5953a450e95579e87acefeecfc5b/source/ide.js#L1699-L1704)). You can't add properties to elements using the second parameter. Instead, you'd have to add them manually like this:

```js
const fileImport = document.createElement("input");
fileImport.type = "file";
fileImport.style.display = "none";
fileImport.accept = ".tscipt";
```

distribution/tscript.js
```diff
@@ -19306,7 +19306,25 @@ function fileDlg(title, filename, allowNewFilename, onOkay)
...
 
 	function importFile(){
-		alert('Import');
+		// <input id="file-import" type="file" style="display: none;" accept=".tscript">
+		let fileImport = document.createElement("input", {
+			"id": "file-import",
+			"type": "file",
+			"style": "display: none;",
+			"accept": ".tscript"
+		});
```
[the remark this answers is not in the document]
TGlas: You can also use the convenience function `tgui.createElement`, which is much closer to your approach.

TGlas:
Great addition! Please fix the issues pointed out by Jartreg, then I will merge this into the project.
Please do changes only to the files in "source". You can test them by opening the index.html file in that directory. The distribution files are overwritten automatically by the makefile. I can do that after merging.

Tilo-K:
I'm still getting the broken link error for the examples. But I currently don't know how to fix this because the `source/def-examples.js` is not included in the repository. If I should be able to fix this please let me know.

TGlas:
The problem with the missing examples file was on my side; I had apparently forgotten to add it to the repository. Please go a pull, should be there now.

PR #40, opened 2020-12-01
fixed missing return value in core function 'assert'

sorachan:
Fixes bug that caused an internal error when `assert` statements were evaluated such as in the common typo `assert(x == 0):`


TGlas:
Thanks a lot!

Issue #41, opened 2020-12-01
Dialogs could be improved


manuel-fischer:
![grafik](https://user-images.githubusercontent.com/58823087/100776319-af454c80-3404-11eb-82ae-1c3c03e9cd56.png)

On the left there is the current load file dialog, on the right there is my improved version that I drew. I will describe the changes and the issues the current gui has.

- The problem with the current buttons
  - inconsistency: some buttons show the hand cursor([Delete file] etc...) others show the normal arrow ([Okay] and [Cancel])
  - either the buttons could be changed to be 3D or to go with the modern look, the look could change like the following
    - this would give the user feedback that there is a button that is clickable
    - 1: normal state, 2: mouse over state, 3: mouse button down state
![grafik](https://user-images.githubusercontent.com/58823087/100778187-fcc2b900-3406-11eb-8790-d5685ca529b5.png)
- X-button to close the dialog: its sometimes hard to directly see the cancel button (file dialogs) or the close button (configuration), it might be useful if there was an additional X-button in the titlebar of the dialog
- Arrangement of buttons in the file dialogs
  - I suggest to only keep [Okay] and [Close] at the bottom of the dialog and move the other three buttons above the file list, this would distinguish the two types of buttons:
    - [Okay], [Close]: buttons that close the dialog
    - [Delete], [Import], [Export]: that operate on the files in the list and don't close the dialog
  - Toolbar:  [Delete], [Import], [Export] can also be icon buttons.
    - Status bar: shows information about all files or the currently selected file (number of lines, date, etc.) specifically
- Double click on list element in file dialogs should confirm the selected file
- Also add text box to the load file dialog, it might be useful to search for a file, when there are many files
  - when there is text in the text box, elements in the list, that contain the text, could be filtered or highlighted
  - the best match could be selected

- I also painted a Export all button, that might be helpful when creating backups of the files.
  - it might also export all the json data stored by the TScript function `save`

# Edit
- The [Import] button closes the dialog but does not open the imported file when it was imported.
  - I suggesst that the dialog keeps open, so other files can also be imported, and the newly imported file is selected

I am working on this in my branch [gui-improvements](https://github.com/manuel-fischer/tscript/tree/gui-improvements)

TGlas:
I totally agree. The dialogs started out with minimal functionality, rather quick and dirty. They never got much attention. Then more functionality was added over time, without any guidelines. Doing a systematic rehersal is a good idea, since usability is important. I am looking forward to the improved version!

[...]

PR #45, opened 2020-12-02
Gui improvements

manuel-fischer:
There are some internal changes but the main changes should be shown by the following screenshots:

grafik

grafik


manuel-fischer:
Reworked set hotkey dialog. It shows more information for the current button, like the tooltip. And you can also see that some tooltips got shortened: "run the program or continue running the program" -> "run/continue program"

grafik

manuel-fischer:
A problem currently see, is that some buttons on the toolbar get invisible, when the viewport size is less than around 1140 px, I think about changing the documentation button to an icon button and making the program status and minimized panel bar flexible in width

manuel-fischer:
grafik

Small width of viewport, spacing gets minimized, status box gets shortened. Documentation button always stays on top, never wraps around

There is a bug, that some panels are clipped at the right.

TGlas:
This looks really good. I wonder what further plans you have for it.

manuel-fischer:
Currently I only have the improvement ideas below, but if someone comes up with an improvement idea, I could also implement it here. I will keep this draft pull request a little bit open.

1. Improving the search bar of the editor, by adding buttons to it and an additional button to the toolbar, that opens the search bar. Possibly adding such a search bar to the stack, program and messages panel (one could argue that it might not be necessary because the browser itself also has a search bar):
grafik
2. Adding line number to stack frames in the stack panel, clicking on the entry goes to the corresponding line in the editor
3. Pressing F1 opens the documentation: Problem: Browsers like Firefox sometimes disallow opening an additional tab when pressing a key. I already added the functionality, that if you have something in the editor selected, then the selection is searched in the documentation (only keeping alphabetic characters). It might be useful to directly be able to press F1 in the editor. This could be enabled by adding additional documentation panel(s), that can be opened in the same browser tab. So pressing F1 would open the documentation in the IDE and clicking on the documentation button would open it in another tab. An additional button could also allow it to open the documentation in the IDE.

Jartreg:
I think something that should be improved is *how* CSS is used in the IDE. In many places, elements are repeatedly given the same inline styles in JS, although it could be solved much easier with CSS classes. In addition, the layout of the IDE relies too much on absolute positioning and `calc(...)`. Many of their use cases could (and IMHO should) be replaced with other layout methods, which would also bring the advantage of being less likely to break if someone gets creative with their window sizes.

The [file dialog](https://github.com/manuel-fischer/tscript/blob/5d19f590349d937c851ae5474d8a8e51ed834a44/source/ide.js#L1586) is a great (or not so great) example for this:
* Buttons are give the same `margin-left` of `10px` as an inline style, which should be better done with a CSS class.
* The file list is positioned absolutely and with `calc(...)`, although something like flexbox would do the same job with less possibilities to break layout.

**However**, if you agree I think it would be better to open an issue instead and work on this in a separate PR once this is done, since it might be out of scope for this PR.

What do you think?

TGlas:
@Jartreg I completely agree that flexbox is underused in the project... I think I did not use it at all at the time. Many layout details were done quick and dirty without even aiming for a proper solution. There is plenty of space for improvements. But I also agree that the technical side of things can be (mostly) separated from look&feel improvements.

manuel-fischer:
I made some changes/refactoring to CSS styles, I already moved some of the styles out of the Javascript code into their own CSS class and I am going to do it for other styles. The Issue I see is that there are two CSS files that both have style classes both used in the IDE.

manuel-fischer:
Here are some screenshots:

grafik
_Added Icons and other changes_

grafik
_Floating panels are now improved and also show a slight shadow beneath them, improving visibility_

manuel-fischer:
I am almost ready with this draft pull request, if there are some other improvements that can be made, let me know. Otherwise I am going to do a final cleanup and I am going to make it ready for review.

manuel-fischer:
Very large icons

grafik

manuel-fischer:
I am almost ready with this draft pull request, if there are some other improvements that can be made, let me know. Otherwise I am going to do a final cleanup and I am going to make it ready for review.

TGlas:
This is already a herioc effort -- something really important, and something nearly nobody considers a rewarding job. Just call it good and let's move on to the code review.

distribution/tscript.js
```diff
@@ -7785,6 +7785,45 @@ module.createLabel = function(description)
...
+		"type": "canvas", 
+		"parent": description.parent,
+		"classname": "tgui", 
+		"style": style,
+	});
+	canvas.width = description.width;
+	canvas.height = description.height;
+	
+	// zoom
+	// TODO: programmaticly detect whenever zoom changes
```
Jartreg, line 7814: I don't think browser zoom is something that needs be handled in such detail. It can't even be guaranteed to work the same in each browser. When taking a look at macOs, for example, Safari, Firefox and Chrome do pixel ratio zooming *when using keyboard shortcuts* (Firefox also when using the scroll wheel). However, *when using the trackpad to zoom*, all three use a different method of zooming, which still results in pixelated icons:

trackpad zooming in Safari
*This is the ide in Safari, zoomed to its maximum using the trackpad. Doing this doesn't change the `devicePixelRatio` in either browser.*

To really get the icons to scale properly, it would be necessary to use SVG instead of Canvas. This would however require a rewrite of each icon. For now, the current method (static handling of `devicePixelRatio`) should be fine.

TGlas: That' true, browsers (in particular mobile browsers) handle zoom very differently, and often inconsistently. In my experience, having a higher (say, doubled) resolution and letting the browser and the user do whatever zooming they deem appropriate is a good-enough solution, and definitely far better than assuming that one canvas pixel maps to one device pixel. I have done this with large documents with very many pixels (several A4 pages at 300 dpi) running into hitting performance problems, so it should be feasible for icons to have a quite high resolution.

manuel-fischer: The problem is that icons keep the resolution after they are created. Reloading the website while zoomed in would update the icons with the new resolution.

There are the following ways that zooming icons could work:

1. keep small 20x20 px icons at all zoom levels
2. use 40x40 px icons at all zoom levels, the difference to 20x20 px is barely noticeable at 100% zoom level. For larger zooms it looks definitely better than 20x20
3. keep it as it currently is: use `devicePixelRatio`
  a) use the current `devicePixelRatio` whenever an icon is created -- current
  b) use the initial `devicePixelRatio`, when the page was loaded
  c) try to keep icons up to date with the `devicePixelRatio`, whenever it has changed
4. add an option to the configuration to allow users change the icon size
5. use SVG icons
  a) embedded into the website
  b) created at runtime
  c.) ~external files~

What do you think, which is the best option?

I also thought about adding scaling to the canvas, it might look nicer when the website is zoomed in, but it might also reduce performance

Jartreg: As already said: I *really* don't consider this to be a major issue (correct me if you think otherwise). Because of that, let's just leave it as it is and go with 3a for now. *In the future*, we could try to switch to SVG, but I think it's **out of scope for this PR**. I could however create an issue as a reminder.

Jartreg:
closes #41

Jartreg:
@TGlas it seems that I don't have the permissions to link this PR with #41 but it might be a good idea to close the issue when this is merged.

distribution/tscript.js
```diff
@@ -7812,7 +7812,8 @@ module.createCanvasIcon = function(description)
 	
 	// zoom
 	// TODO: programmaticly detect whenever zoom changes
-	let zoom = window.devicePixelRatio;
+	let zoom = 2; // Good enough
```
Jartreg, line 7815: Why wasn't `window.devicePixelRatio` good enough? Didn't you want to be flexible with scaling?

distribution/tscript.js
manuel-fischer: I think that a fixed zoom of 200% is good enough. It shouldn't be any problem when the page is not zoomed in, as it usual is. But when it is zoomed in, the zoomed in icons look nicer without the need of updating. Note that with the `devicePixelRatio` uneven zooms like `150%` are blurry, because lines are rendered at subpixel level. Shrinking `200%` icons to those zooms looks almost the same.

distribution/tscript.js
manuel-fischer: see other conversation

distribution/tscript.js
Jartreg: While I agree that 200% is good enough, you might be misunderstanding `devicePixelRatio`. It is not supposed to be an indicator of the zoom level but rather the scaling factor of screens. Most current laptop screens (mine included) and many external monitors already have a scaling factor of `2` when completely zoomed out:

Chrome Devtools showing devicePixelRatio being 2

Using `devicePixelRatio` then is rather a matter of adapting to *screens* instead of *zoom levels*.

manuel-fischer:
Some icons changed a little bit (mainly in https://github.com/TGlas/tscript/pull/45/commits/f289216e55c5d32db5e6e3addb17641c3b8e2e8a)

grafik

Those and others (save as, breakpoint, minimize, export) were tweaked a little bit

Note that 1px was added between minimized icons

distribution/tscript.js
manuel-fischer: Technically `devicePixelRatio` is the zoom level on the pixel level. This is what I meant with zoom. It determines to how many physical pixels on the screen a length of `1px` in CSS style sheets correspond to. Note: you might have a 4k monitor but a logical screen size of 1080p, websites and other applications work as if your screen had 1080p, only when rendering at the pixel level every pixel unit is scaled by 2. When zooming in a browser it is meant to change the logical screen size of the website relative to the system logical screen size. So zooming 200% in the browser is physically a zoom of 400% (on the pixel level), this physical value is reflected by `devicePixelRatio`.

Jartreg: Okay we seem to mean exactly the same thing, but I still don‘t understand why using a static factor of 2 would be better than using `devicePixelRatio`. If you want to give the icons a higher resolution to account for zooming, people who already have a `devicePixelRatio` of 2 because of their screen don‘t get any benefit at all. Instead, just using `devicePixelRatio` would assure that icons look good in the _default_, zoomed out state for _everyone_. Antialiasing will most likely happen anyway, and is highly dependent on the browser engine. Just that is IMO not enough to dismiss `devicePixelRatio`. If you still want to avoid it, you can still [disable it](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/imageSmoothingEnabled).

TGlas:
> @TGlas it seems that I don't have the permissions to link this PR with #41 but it might be a good idea to close the issue when this is merged.

The link is not established.

distribution/tscript.js
Jartreg: Since this seems to be going nowhere and is not a real issue for me, let's go with you default of 2.

Jartreg (approved):

source/codemirror-tscriptmode.js
```diff
@@ -19,6 +19,7 @@ CodeMirror.defineMode("text/tscript", function(config, parserConfig)
 			"then": true,
 			"else": true,
 			"for": true,
+			"in": true,
```
TGlas, line 22: I see why you want to do that, but `in` is not a keyword (I changed that at some point). For example, a variable can be called in:

    var in = 7;
    print(in);

That's why I think that this highlighting would be confusing. Alternatively, we could discuss whether `in` should become a keyword or not.

TGlas (changes requested):
I have only one detail on syntax highlighting ("in" is not a keyword), the rest looks good to me.
Please remove the zip file, I see no need to keep it in the repo, and it causes unnecessary conflicts.

source/codemirror-tscriptmode.js
manuel-fischer: I think that `in` might be useful to be considered a keyword, because the `for`-syntax uses it as a syntactic element. I added `in` to the highlighting previously, because `in` might be highlighted like a keyword in a `for`-loop. `in` might also be considered as the operator to check if an element is in a container, similar to python. But this is not really needed, because `Array` and `Dictionary` have the function `has`, that already is very concise.

TGlas (approved):

TGlas:
I have no idea what happened. I merged the pull request into master using the command line and manual conflict resolution. As far as I can tell, all GUI changes took effect ([https://tglas.github.io/tscript/distribution/index.html](https://tglas.github.io/tscript/distribution/index.html)). However, the PR is not marked as merged. Does anyone know how to debug this?

Jartreg:
If you take a look at b21ee2d, you can see that it is not a merge commit, because it only has one parent (9bd38a4, which was master before). Therefore GitHub could not detect it as a merge of this PR. It would be safe to just close this now.

TGlas:
Okay, github let me redo things with the web editor (changing things into themselves), but that seems to resolve the broken state.
Anyway, thanks a lot for this PR and for the discussions. The GUI looks much better now!

PR #46, opened 2021-01-04
Simple audio support

TheHllm:


TGlas:
If I understand this correctly then the sound library would allow to play sound files, for example, hosted on a public website.

I thought about audio support of various types (including embedded and programmatically generated waveforms) several times. It would be a great feature for games. I rejected the idea for two main reasons:
- I'd like to avoid dependencies on "resources" (external files) like sound and image files. TScript programs should be self-contained. That makes it difficult to do something meaningful with sound (and other media like images and videos, which are all directly supported by the Browser).
- I want TScript to be embeddable into applications. The current E-Learning system is just a first step. An even more important step that will follow hopefully soon is an electronic exam system (everything is ready from my side, just waiting for university facilities). In such an environment you don't want dependencies on externally hosted resources, and for sure you don't want the ability to play sound. That could of course be overruled by other means, but still, it would be in the way.

TheHllm:
> If I understand this correctly then the sound library would allow to play sound files, for example, hosted on a public website.
Yes.

PR #50, opened 2021-01-07
Audio playback

TheHllm:
Allows the playback of locally generated buffers.
Example code and a basic how-to can be found in the docs.


source/tscript.js
```diff
@@ -2112,6 +2112,106 @@ let lib_canvas = {
...
+			"audio":{
+				"MonoAudio":{
+					"constructor":function(object, buffer, sampleRate){
+						if (! module.isDerivedFrom(buffer.type, module.typeid_array)) this.error("/argument-mismatch/am-1", ["buffer", "audio.MonoAudio.constructor", "array", module.displayname(buffer)]);
+						if(! module.isDerivedFrom(sampleRate.type, module.typeid_integer)) this.error("/argument-mismatch/am-1", ["sampleRate", "audio.MonoAudio.constructor", "integer", module.displayname(sampleRate)])	
+						let buf = this.service.audioContext.createBuffer(1, buffer.value.b.length, sampleRate.value.b);
+						let ar = buf.getChannelData(0);
+						//copy from buffer to js objects buffer
+						for(let i=0; i < ar.length; i++){
+							//maybe check if number is a real and between -1 and 1?
```
TGlas, line 2147: Yes, please! I actually don't know what happens if we give JS a junk value at this point (I assume that the audio buffer is a typed array), so the user should get an error message if values are non-numeric or NaN. On the other hand, for the values I think that clipping to [-1, 1] would be more reasonable than raising an error.

source/def-stdlib.js
```diff
@@ -678,5 +678,65 @@ if (doc) doc.children.push({
...
+
+				var l_sf = (l_freq / sampleRate) * pi(); 
+				var r_sf = (r_freq / sampleRate) * pi(); 
+
+				var l_samples = [];
+				var r_samples = [];
+
+				var c = 0;
+				for var i in 0:(sampleRate*lenInSeconds) do{
+					l_samples.push(sin(c));
```
TGlas, line 724: I find sin(l_sf * i) more natural, and I would even go for sin(2 * math.pi() * l_sf * i), since that would turn l_sf into a "proper" frequency (measured in Hz).

examples/audio.tscript
```diff
@@ -0,0 +1,28 @@
...
+
+var l_sf = (l_freq / sampleRate) * pi(); 
+var r_sf = (r_freq / sampleRate) * pi(); 
+
+var l_samples = [];
+var r_samples = [];
+
+var c = 0;
+for var i in 0:(sampleRate*lenInSeconds) do{
+	l_samples.push(sin(c));
```
TGlas, line 17: see below comment on the documentation

TGlas (changes requested):
Looks good overall, but please fix the points addressed in the comments.

source/tscript.js
TheHllm: The array is typed and JS will convert anything invalid to NaN.
image

source/def-stdlib.js
TheHllm: `l_freq` was supposed to be the frequency in Hz and  `l_sf` is the 'sample frequency' (i.e. the frequency per sample, not to be confused with sample rate). But i can see how this would be confusing as an example.

source/tscript.js
```diff
@@ -2167,26 +2187,21 @@ let lib_canvas = {
...
+					"constructor":function(object, leftBuffer, rightBuffer, sampleRate, performTypeCheck){
 						if (! module.isDerivedFrom(leftBuffer.type, module.typeid_array)) this.error("/argument-mismatch/am-1", ["leftBuffer", "audio.StereoAudio.constructor", "array", module.displayname(leftBuffer)]);
 						if (! module.isDerivedFrom(rightBuffer.type, module.typeid_array)) this.error("/argument-mismatch/am-1", ["rightBuffer", "audio.StereoAudio.constructor", "array", module.displayname(rightBuffer)]);
-						
+
+						if(! module.isDerivedFrom(performTypeCheck.type, module.typeid_boolean)) this.error("/argument-mismatch/am-1", ["performTypeCheck", "audio.MonoAudio.constructor", "boolean", module.displayname(performTypeCheck)])	
 						if(! module.isDerivedFrom(sampleRate.type, module.typeid_integer)) this.error("/argument-mismatch/am-1", ["sampleRate", "audio.StereoAudio.constructor", "integer", module.displayname(sampleRate)])	
-						debugger;
+						
 						let buf = this.service.audioContext.createBuffer(2, leftBuffer.value.b.length, sampleRate.value.b);
```
TGlas, line 2197: Please simply return here if this.service.audioContext is null or undefined. To me, the "service" interface is completely optional. E.g., if sound is not available for some reason, then the "service" is null. That may happen in particular in contexts without GUI, e.g., when automatically checking program code. Then the call should still succeed, but no JS audio object is ever created. The same applies to the play() method. Sorry, I should have noticed that earlier.

source/def-stdlib.js
```diff
@@ -694,6 +694,9 @@ if (doc) doc.children.push({
 				<code class="code">StereoAudio(leftBuffer, rightBuffer, sampleRate)</code> provides the same functionality as above, but allows playback of stereo audio.
 			</td></tr>
 			</table>
+			<p> Both constructors allow disabling checks that ensure the arrays only contain floats and are in the proper ranges. 
+			This is disabled by default and can be set by calling the constructor with an optional trailing <code class="code">false</code>.</p>
+			
 			<h3>Functions</h3>
 			<tscript>
 			function play(){}
```
TGlas, line 702: The documentation should mention that sound is playes asynchronously, i.e., in the background, while the call returns immedately and the program continues to run. The alternative would be to wait for the sound to complete. Sure, that's rarely what's needed, but to a beginner the behavior may still be surprising.

TGlas (changes requested):
getting closer :)

Jartreg (changes requested):
It might be better to align the code style of the examples with the rest of the documentation, to be consistent. These `for`-statements, for example, use a different indentation for the opening brace than the rest of the documentation ([Example: for-loops in the documentation](https://tglas.github.io/tscript/distribution/index.html?doc#/language/statements/for-loops)) .

examples/audio.tscript
```diff
@@ -0,0 +1,26 @@
...
+var durationInSeconds = 2;
+var l_freq = 440; # frequency in Hz 
+var r_freq = 554;
+var sampleRate = 48000; # sampleRate in Hz
+
+
+var l_samples = [];
+var r_samples = [];
+
+for var i in 0:(sampleRate*durationInSeconds) do{
```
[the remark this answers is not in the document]
TheHllm: Shure why not is there any way for me to accept all theses changes at once?

source/def-stdlib.js
```diff
@@ -720,21 +720,23 @@ if (doc) doc.children.push({
 				var l_samples = [];
 				var r_samples = [];
 
-				for var i in 0:(sampleRate*durationInSeconds) do{
+				for var i in 0:(sampleRate * durationInSeconds) do
+				{
 					l_samples.push(sin( i * (l_freq / sampleRate) * pi()));
```
TGlas, line 725: missing 2 *

TGlas (changes requested):
Only a tiny detail left! Please fix, then I'll merge the branch.

TGlas:
Thanks a lot for the effort!

PR #52, opened 2021-01-14
Impoved performance

TheHllm:
This PR allows tscript to utilize all of the performance based on device and current load, by using `window.requestIdleCallback`.
For calculating the first 3000 prime numbers this resulted in an 57% performance increase on my machine.
This would likely be the opposite on phones etc. but would allow the UI to remain usable.


Jartreg (changes requested):
I like the idea to replace `setTimeout`, but I think there are some issues with `requestIdleCallback`:

### Browser support
`requestIdleCallback` is not yet supported by Safari ([see support tables on MSN](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback#browser_compatibility)), so you would need to add a fallback that case.

### Background Tasks
For calculating prime numbers it is completely appropriate to use background tasks. However, when using the canvas, for example, the code needs to be sure to be run on every frame, and `requestIdleCallback` is not designed for that case.

## Suggestion: use requestAnimationFrame instead
I think your performance improvements came from the better scheduling of `requestIdleCallback`. To guarantee that also canvas-based scripts benefit from this, I would suggest to try [`requestAnimationFrame`](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame) instead. If you see similar performance improvements, it might be better to use that, otherwise you should add a timeout to your calls to `rIC` and use `rAF` as a fallback.

TGlas:
Interesting approach, I did not know about `requestIdleCallback`. Although I am well aware of the fact that my `setTimeout` loop is an ugly and somewhat outdated design.

Jartreg is of course right that a fallback solution is needed as long as this is not supported in all major browsers, and you should also test how a demanding canvas application performs with your method. `requestAnimationFrame` is surely also an interesting method, in particular for canvas applications, but I don't think that it really solves the problem of using all available CPU time. The current scheduler is designed to leave a certain fraction of the time (roughly 10%) to the browser so it stays reactive. However, I am impressed that you can get such a big speedup.

In the end, I think that the speedup should not be the main reason for doing this. There are surely much more substantial ways to accelerate execution, like compiling (parts of) the program to Javascript. The scheduling should primarily be safe and sound. For the moment, `setTimtout` does the job, but I am open to improvements, as long as they work everywhere.

Jartreg:
@TheHllm could you share the exact code that you used to test performance?

TheHllm:
@Jartreg 
When testing be sure to only have one tab open.
```
for var qqq in 0:5 do{
var start=localtime();

var primes = [];
primes.push(2);
for var i in 3:3000 do{	
	var isPrime = true;
	for var j in 0:primes.size() do {
		if(i % primes[j] == 0) then{
			isPrime = false;	
		}
	}
	if(isPrime) then {
		primes.push(i);
	}
}

var end = localtime();
print(end-start);
}

```

Jartreg:
So I've done some testing, and from what I've found out I don't think that `requestIdleCallback` is a good idea. I used different code for calculating prime numbers, because I had started before you sent it.

<details>
<summary>Test Code</summary>

```js
function isPrime(n, primes) {
	for var e in primes do {
		if e * e > n then break;
		if n % e == 0 then return false;
	}
	return true;
}

function findNPrimes(n) {
	var primes = [];
	var i = 2;
	
	while primes.size() < n do {
		if isPrime(i, primes) then primes.push(i);
		i += 1;
	}
}

var start = time();
findNPrimes(3000);
print(time() - start);
```

</details>

_This has become far longer than it should have... 
**TL;DR** `rIC` is only any good in Chrome and completely destroys canvas_

## What should we optimze for?
When thinking about what TScript can mainly be used for, two things come into my mind:

1. Computing (e.g. our prime number tests)
2. Graphics/Games (e.g. [LatticeCraft](https://github.com/manuel-fischer/LatticeCraft))

<details>
<summary>Computing</summary>

### Computing (primes!)
`requestIdleCallback` seems like a good choice for this, but I compared it with the current and some other methods:

A chart showing the time it took to calculate the first 3000 prime numbers in different browsers

_Times are the average of 10 runs, for each of which the borwser was restarted. Lower is better. (Safari has no bar for `rIC` because it doesn't support it.)_

When looking at the chart, I'm pretty sure that you are using Chrome, because it's the only browser that seems to get a significant speedup at all. Also, in contrast to my initial thoughts, `requestAnimationFrame` only gets better results in Firefox.

So while it's nice that Chrome seems to be very good with handling idle time, I don't think worth it to optimize for just one browser. In addition, you can see that other methods I tried (green and orange) also achieved significant improvements across all browsers.

What I did there is I increased the time budget for the `chunk` function from 14 to 30 milliseconds. This allows TScript to run more code in one frame while allowing the browser to keep a healthy 30FPS. When writing computing-focused programmes, I think users can expect the performance to be impacted by that. Allowing the frame rate to drop to 30 FPS keeps the page well usable while giving more time to the code. It also works with graphics-based applications, since in most cases, they yield back to the browser quickly enough to achieve 60 FPS.

But I'm not done yet...

</details>
<details>
<summary>Graphics/Games</summary>

### Graphics/Games
To test how graphics-based programmes perform, I bodged together something that shows how often the `timer` event gets emitted per second (which is basically the frame rate then) while using the canvas to a certain extent: [See Gist](https://gist.github.com/Jartreg/d39a85adb4c1a23f336620d6dda5b128)
_Psst! If you want to see something more graphics-heavy: [here](https://gist.github.com/Jartreg/bf1be93aefd127397d87a772bcdbc68c) (just don't look at the code mess)_

There's actually not much to say about `setTimeout` and `requestAnimationFrame`, except that in Safari, when you are unplugged from the power supply and the canvas is focused, the frame rate drops by 10 when using `setTimeout`. After clicking out of the canvas, it rises to ≈60 again 🤷. **And then I tried Chrome...**

As noticed before, Chrome is really good with scheduling idle time, but it also seems to make good use of delaying idle callbacks when it doesn't have the time. When running the above test, I only get a frame count of ≈20. To make it worse, when scrolling the in the code editor or on the canvas itself, it freezes completely.

So while `requestIdleCallback` is good for background tasks (as intended), it seems to make the graphics side of TScript a lot less enjoyable.

---

</details>

**&rarr;** I would propose to close this for now, and maybe take a look at the other methods.

TheHllm:
Agreed. I did not think that it could have such an impact on the canvas side of things.

PR #54, opened 2021-01-15
Fix use of undefined AudioContext in WebKit

### LINKED ISSUE #53: [BUG] The interpreter doesn't run in Safari because of AudioContext support
When trying to run code in the IDE in Safari, nothing happens.

**This should be considered critical, since it _prevents any code from running_.**

### Cause

The console shows the following error:

```
ReferenceError: Can't find variable: AudioContext
```

This means the reason for this bug is the audio API added in #50, which tries to create an instance of `AudioContext` without checking whether it is supported ([here](https://github.com/TGlas/tscript/blob/master/source/tscript.js#L6969)). Because in Safari it's [only supported with a prefix](https://developer.mozilla.org/en-US/docs/Web/API/AudioContext/AudioContext#browser_compatibility), this leads to an error that prevents the interpreter from starting.

Jartreg:
fixes #53

### DIFF
--- a/distribution/index.html
+++ b/distribution/index.html
@@ -7344,7 +7344,10 @@ module.Interpreter = function(program)
 		}
 	};
 
-	this.service.audioContext = new AudioContext();
+	var AudioContext = window.AudioContext || window.webkitAudioContext;
+	if (typeof AudioContext === "function") {
+		this.service.audioContext = new AudioContext();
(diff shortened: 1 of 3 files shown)


TGlas:
Safari is well known for lagging behind multimedia support in many ways. Thanks for the fix!

PR #57, opened 2021-01-22
Resolves #55

### LINKED ISSUE #55: Refactoring
The code **badly** needs refactoring.
Some things that i propose need changing:

- not having a distribution folder / using ci to update a live version on e.g. an other github repositry
- proper build process (move to [typescript](https://stackoverflow.com/a/58042841/8055836))
- separating components into files (move to typescript)
- separating files into folders (for codemirror etc.)

But with 345Kb of js in a single file implementing any of these would likely require more than one person and a consensus on approaching this.

TheHllm:
Resolves #55

# Features
- Typescript and webpack as bundler and builder
- Refactored code into individual files
- Automatic deployment to github pages
- Automatic unittest on push and on pull requests.
- `localStorage` is now a service and can be used in tests

- [Live Example](https://thehllm.github.io/tscript/distribution) 
- ~~[Tests](https://thehllm.github.io/tscript/distribution?unittest)~~

When merged set github pages to use the `dist` branch.


TGlas:
Wow!

This looks like a lot of work. I did not look at all the details yet, but let me comment your summary.

- Typescript and webpack as bundler and builder: I see that a solution like webpack makes a lot of sense for the project, in particular when splitting the main file into manageable chunks. However, I don't understand the role of typescript in the process. 
- Refactored code into individual files: As indicated above, very useful for further development.
- Automatic deployment to github pages: Nice feature.
- Automatic unittest on push and on pull requests: Extremely useful!
- localStorage is now a service and can be used in tests: Does that mean that it becomes is a "service" of the interpreter? So it can be cleared after each unit test... that would of course make sense.

My main question here really is about the roles of webpack and typescript, and whether they can be separated. I don't have hands-on experience with any of them, but I am eager to learn.

TheHllm:
# Webpack and typescript
As you are probably aware typescript is a superset of javascript. That means that i could reuse ([example](https://github.com/TheHllm/tscript/blob/d442d37bb6f1cb015fdadaf14fc3843afb2b130e/source/src/gui/codemirror-tscriptmode.ts)) a lot of code. Typescript however also allows static checking of types, wich was very useful when for example making the `options` (i.e. `checkstyle` etc.) part of each [parsed programm](https://github.com/TheHllm/tscript/blob/d442d37bb6f1cb015fdadaf14fc3843afb2b130e/source/src/parse/parser.ts#L15).
The use of typescript is however completely optional for webpack. Webpack is only responsible for [linking ](https://github.com/TheHllm/tscript/blob/d442d37bb6f1cb015fdadaf14fc3843afb2b130e/source/src/parse/parser.ts#L1-L12)(somewhat like the c++ linker would) the files together and outputing a file that contains js css and html (these can also be seperate webpack is quite powerfull).
Running [`npm interop`](https://github.com/TheHllm/tscript/blob/d442d37bb6f1cb015fdadaf14fc3843afb2b130e/source/package.json#L9) will generate set of javascript files that [export the Interpreter etc.](https://github.com/TheHllm/tscript/blob/d442d37bb6f1cb015fdadaf14fc3843afb2b130e/source/src/interop.ts) into a file that can be `requiere()`'d by, for example a nodejs application with out the use of [ts-node.](https://www.npmjs.com/package/ts-node).

# localStorage
Yes localStorage is now part of the service property of the interpreter. The unittests are currently set up to use the `window.localStoarge` if it is defined, that way the tests in the browser actually test the underlying infrastructure. When running the tests via mocha (the test framework i used) a [moc localStorage](https://github.com/TheHllm/tscript/blob/d442d37bb6f1cb015fdadaf14fc3843afb2b130e/source/src/interpreter/defaultService.ts#L1-L20) is used.

TGlas:
I had a deeper look into this and also into typescript, and I must say that I like the approach a lot! The project is so tidy and well organized, it's a real joy to navigate. So, I would like to merge this PR and move on with typescript. One thing, and that may actually be quite some work, is that in the meantime the GUI improvements PR was merged, resulting in lots and lots of conflicts. Do you see a good way to resolve them before merging?

.github/workflows/cd.yml
```diff
@@ -0,0 +1,47 @@
...
+        node-version: ${{ matrix.node-version }}
+    - run: npm ci
+      working-directory: source
+    - run: npm run build
+      working-directory: source
+    - run: npm test
+      working-directory: source
+
+# Push dist/ to dist
+    - run: git checkout --orphan dist
```
Jartreg, line 35: Why not `gh-pages`? I think this is the default, and as long as we don't have any other purposes for this branch, `gh-pages` would be the most descriptive.

Jartreg (changes requested):
I regret that I couldn't convince you to do this in smaller steps and discuss this more before implementing, but I guess if it works then it's fine... 🤷

I haven't looked much into it yet but I see some problems already:

- [ ] ~~The CI and CD workflows will not work. They assume `package.json` to be in the repo root but it's actually in `source`.~~
- [x] The `tsconfig.json` might need some attention. It also shows an error before installing dependencies, which might be intimidating to new contributors. Imagine having errors before even touching anything.
- [x] Deploying the distribution artifacts to the branch root is logical, but it will break the page for all users which assume it to be at `/distribution/index.html`. This will be everyone who has made a bookmark, uses URL bar suggestions or comes from a page without updated links.
- [x] `contributing.md` should be at the repo root to be better recognizable.

A few additional things/suggestions:

- [x] I think it would be better to move everything from `source` into the repository root. This would also fix the problems with CI/CD and would allow NPM commands to be run withoud changing directories. Having a directory called `src` inside `source` seems unnecessary. Also, `css` seems like it belongs in `src`.
- [x] `TODO.md` should be an issue.
- [x] `description.txt` files should be named `README.md`. This would make them appear in the GitHub directory view.
- [x] `/dist` could be ignored entirely.
- [x] Creating _one_ `.gitignore` at the root might be better.
- I would have strongly suggested to go with something other than webpack, but it seems to work well so it should be fine.
- The title of this PR is not very descriptive.
- ~~If you want to close `#55` with this, please use the proper keywords and do so in the description or a comment.~~

Jartreg:
For the transition, please remember that in the time frame between merging this and @TGlas changing the GH Pages branch to `dist`, the page will not be accessible. To solve this, maybe `distribution` should still be included here and only removed after the `dist` branch is ready and deployed.

.github/workflows/cd.yml
```diff
@@ -0,0 +1,47 @@
...
+      with:
+        persist-credentials: false
+        fetch-depth: 0
+
+# Compile and run tests
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: npm ci
```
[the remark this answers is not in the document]
TheHllm: The workflows works just file.
https://github.com/TheHllm/tscript/runs/1751299306

source/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "css-loader": "^5.0.1",
+    "html-inline-css-webpack-plugin": "^1.10.0",
+    "html-inline-script-webpack-plugin": "^1.0.1",
+    "html-webpack-plugin": "^5.0.0-beta.6",
+    "mini-css-extract-plugin": "^1.3.4",
+    "mocha": "^8.2.1",
+    "ts-loader": "^8.0.14",
+    "ts-node": "^9.1.1",
+    "typescript": "^4.1.3",
+    "uglifyjs-webpack-plugin": "^2.2.0",
```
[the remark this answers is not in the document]
TheHllm: - ts-node is needed for mocha
- typescript for `npm run interop` (transpiling for use in grading system)
- ts-loader is used by webpack

But yes uglifyjs can be remove.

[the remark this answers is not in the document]
Jartreg: Sorry I actually commented only on line 27. It seems that GitHub includes a couple of lines before that too...

.github/workflows/cd.yml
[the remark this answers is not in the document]
Jartreg: Ah sorry, I totally missed the `working-directory: source` below and wondered why it did't work in the terminal... Could you mark this and https://github.com/TGlas/tscript/pull/57#discussion_r563279923 as resolved?

source/webpack.config.js
```diff
@@ -0,0 +1,45 @@
+const path = require('path');
+const LoaderOptionsPlugin = require('webpack/lib/LoaderOptionsPlugin');
+const MiniCssExtractPlugin = require('mini-css-extract-plugin');
+const UglifyJsPlugin = require('uglifyjs-webpack-plugin');
```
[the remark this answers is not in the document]
TheHllm: MiniCssExtractPlugin is used.
The rest is now removed

source/contributing.md
```diff
@@ -0,0 +1,7 @@
+# Development Environment
+- install nodejs 
+- install npm
+- run `npm ci`
+
+# Compiling
+To compile and continuously check for changes in the source run: `npm run dev`.
```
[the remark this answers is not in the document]
TheHllm: I added a bit more info, but i realy don't know what to write....

source/tsconfig.json
```diff
@@ -0,0 +1,27 @@
...
+        "incremental": false,
+
+        //"strictNullChecks": true //would be nice if all these would work.
+        //"noImplicitAny": true,
+        //"noImplicitReturns" : true,
+        //"noImplicitThis": true,
+        
+    },
+    
+    "include": ["./node_modules/"]
```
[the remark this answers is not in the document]
TheHllm: I did not have any errors. Only when running `npm ci` but these now gone (from removing uglifyjs).

TheHllm:
@TGlas The gui changes are all ported over. As well as the style improvments. The branch `gh-pages` now contains the distribution folder.

TheHllm:
@Jartreg i wanted to do smaller steps, but the problem is that in order to do anything like using typescript or webpack you need to change quit a lot in order to have the code still work. 
At some point it just seemed better to to go in and do as much as i have done.

TGlas:
>     * I would have strongly suggested to go with something other than webpack, but it seems to work well so it should be fine.
Which tool do you prefer? There seem to exist quite a few alternatives, but I have no idea which one to prefer for which reasons. For me, simplicity would be a good argument, i.e., picking something that serves the needs of the project without doing much more at the cost of unneeded complexity (e.g., boiler plate, complicated config files).

Jartreg:
Personally I prefer Rollup, because compared to webpack its config files can be _much much_ simpler than webpack's and it supports [ES Modules as an output format](https://bundlers.tooling.report/non-js-resources/html/inlining-scripts/). What makes it so much simpler in my opinion is that there is no split between loaders and plugins. Not having to configure which loader loads what kind of file makes it much shorter.

**However**, webpack might be the right choice _for now_, as Rollup doesn't support [inlining scripts into HTML](https://bundlers.tooling.report/non-js-resources/html/inlining-scripts/). This would break the advantage that the IDE can be downloaded as a single file.

There are also things like Parcel, which I haven't looked into at all, and esbuild, which looks very promising but misses some features.

src/index.ts
```diff
@@ -0,0 +1,42 @@
+import * as _ from 'lodash';
+import { ide } from "./gui/ide";
+import { doc } from "./doc/doc";
+import { runTests } from "./tests/runtests_gui";
+import { handleCanvas, handleTurtle } from "./gui/standalone";
+
+
```
Jartreg, line 7:
```suggestion
import "./css/ide.css";
import "./css/tgui.css";
import "./css/codemirror.css";
import "./css/documentation.css";

```

webpack.config.js
```diff
@@ -0,0 +1,43 @@
+const path = require('path');
+const MiniCssExtractPlugin = require('mini-css-extract-plugin');
+const HtmlWebpackPlugin = require('html-webpack-plugin');
+const HtmlInlineScriptPlugin = require('html-inline-script-webpack-plugin');
+const HTMLInlineCSSWebpackPlugin = require("html-inline-css-webpack-plugin").default;
+
+
+module.exports = {
+  entry: ['./src/index.ts', './src/css/ide.css', './src/css/tgui.css', './src/css/codemirror.css', './src/css/documentation.css'],
```
Jartreg, line 9:
```suggestion
  entry: './src/index.ts',
```

Jartreg (changes requested):
Suggestion: import css files in `index.ts` instead of adding them as entries

package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "tscript",
+  "version": "1.0.0",
+  "description": "",
+  "main": "tscript.js",
+  "scripts": {
+    "build": "webpack --mode=production",
+    "dev": "webpack --mode=development -w -c ./webpack.config.js",
```
Jartreg, line 8:
```suggestion
    "dev": "webpack --mode=development --watch",
```

* `--watch` can be easier recognized than `-w`
* specifying the config file is not necessary

Jartreg (changes requested):
Another suggestion: make the command line for `dev` simpler

Jartreg:
Also something that we shouldn't forget: @TGlas [mentioned](https://github.com/TGlas/tscript/issues/55#issuecomment-761840480) that the interpreter needs to be able to be embedded into the e-learning system _in the browser_. Server-side could be done with the interop build (???), but this won't work in the browser. Maybe this could be solved by adding another bundle that exports the interpreter as `window.TScript` (I don't know how it is used in the e-learning system).

source/tsconfig.json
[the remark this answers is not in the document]
Jartreg: Sorry I wasn't clear: My editor marked the file as having an error, but this is resolved now.

TheHllm:
The interop will also work in the browser.

Jartreg:
When looking at the output, you can see that it uses node's `require` function:

<details>
<summary>Example: <code>distribution/interop/interop.js</code></summary>

```js
"use strict";
exports.__esModule = true;
exports.Tscript_Helpers = exports.Tscript_TypeId = exports.Tscript_Parser = exports.Tscript_Interpreter = void 0;
var typeIds_1 = require("./helpers/typeIds");
var interpreter_1 = require("./interpreter/interpreter");
var parser_1 = require("./parse/parser");
var tscript_1 = require("./tscript");
exports.Tscript_Interpreter = interpreter_1.Interpreter;
exports.Tscript_Parser = parser_1.Parser;
exports.Tscript_TypeId = typeIds_1.Typeid;
exports.Tscript_Helpers = tscript_1.TScript;
```

</details>

This is **not supported by browsers**. Also, the output consists of **50 files**, which will probably take ages to load this way.

TGlas:
> Also something that we shouldn't forget: @TGlas [mentioned](https://github.com/TGlas/tscript/issues/55#issuecomment-761840480) that the interpreter needs to be able to be embedded into the e-learning system _in the browser_. Server-side could be done with the interop build (???), but this won't work in the browser. Maybe this could be solved by adding another bundle that exports the interpreter as `window.TScript` (I don't know how it is used in the e-learning system).

On the client side of the current e-learning system it works as follows. The file `source/tscript.js` is loaded into the website. The file contains only the language, without IDE, docs and all the rest. (The editor is also included, but separately, so that's an independent issue.) Having the language loaded allows the website to run TScript code on the client side. To that end, it registers fake services for its interpreter instance that allow it to record program behavior (print, alert, turtle movement, canvas drawing), all in a quite nasty setTimeout loop. That's for behavior analysis, say, unit tests. In addition, there is a tool that matches abstract syntax tree templates to the parser output in order to enable a simple form of structure analysis.

On the server side things are similar. The "grader" creates a temporary JS file containing the contents of `source/tscript.js` followed by an array of all TScript codes to be run, plus a script that actually runs the TScript codes, collects events, and outputs them in an easy-to-parse format. It then invokes `node` with the temporary javascript file through a `system` call (actually a pipe, since it needs to capture its output). Of course, the TScript interpreter could be loaded as a module instead of being copied verbatim to the temporary file.

So in both cases I need something along the lines of `source/tscript.js` as a target (intermediate target?) of the build process.

TheHllm:
@Jartreg ahh i had it the wrong way round, yes of course the browser cannot use these files. But it would be very simple to compil a file that sets winow.TScript etc.

The only question is should i add this to the webpack config @TGlas?

Jartreg:
Why not add another webpack config? I tried this:

<details>
<summary><code>webpack-interop.config.js</code></summary>

```js
const path = require("path");

module.exports = {
  entry: "./src/tscript.ts",
  module: {
    rules: [
      // compile typescript
      {
        test: /\.tsx?$/i,
        use: "ts-loader",
      }
    ],
  },
  resolve: {
    extensions: [".ts"],
  },
  output: {
    path: path.resolve(__dirname, "./distribution"),
    library: "TScript",
    libraryTarget: "var",
    filename: "tscript.js"
  },
};

```

</details>

What this does is it assigns all the exports from `interop.ts` to a variable called `TScript`. To fully match `source/tscript.js`, the exports of `interop.ts` need to be adjusted.

TGlas:
> The only question is should i add this to the webpack config @TGlas?

Yes, as far as I can tell this would solve the point.

Jartreg:
I don't think having everything in one config is a good idea. When developing, for example, you don't really need the extra bundle. Still having to compile it just slows everything down. It could be better separated by creating a second config file for this.

TheHllm:
tscript.js will now get generated. I tested it by switching it out (into the source of the current master) and running the tests. Every test passes. So it should be a drop in switch.

.github/workflows/cd.yml
```diff
@@ -26,14 +26,15 @@ jobs:
...
+    - run: npm run interop
     - run: npm test
 
 # Push dist/ to dist
     - run: git checkout --orphan dist
     - run: git config --local user.email "f6eae15e-09a6-45d4-b9a1-f694f50b9e21@example.com"
     - run: git config --local user.name "🤖"
     - run: git reset
-    - run: git add -f ./distribution/index.html
+    - run: git add -f ./distribution/index.html ./distribution/tscript.js
```
Jartreg, line 37: But is it really necessary? The IDE includes its own version of the interpreter and the e-learning system as well.

It would be nice if we could separate this from the IDE deployment — @TGlas, would you be fine with running the build yourself for the e-learning system?

TGlas: Short answer: yes.

Running a separate build for this is fine, since the e-learning system is updated rather infrequently. However, I would appreciate to keep the build script inside of the project, since it should always be up to date.

Is it possible to define dependent targets? In a makefile I would define the IDE as the default target, and have a separate target for `tscript.js`. Other targets could be the unit tests and maybe more things in the future, like a documentation self-test. Finally, an "all" target would be nice to have.

TheHllm: You would just need to run `npm run interop` .  In that case there is no point in having it part of the gh-pages branch. But i am for building it anyways to ensure that the build works.

Jartreg: > Other targets could be the unit tests and maybe more things in the future, like a documentation self-test.

NPM scripts (defined [here](https://github.com/TGlas/tscript/blob/cfb3fdafc10fa0f4cfe0bfb3171b79fbf0dccfab/package.json#L6-L11)) basically take the role of make targets. Unit tests, for example, are already included in this. It would be possible to run other scripts from within a script using `npm run a && npm run b` or the package [npm-run-all](https://www.npmjs.com/package/npm-run-all).

TGlas: Thanks for clarifying!

src/index.ts
```diff
@@ -0,0 +1,46 @@
+import * as _ from 'lodash';
```
Jartreg, line 1: This doesn't seem to be used. It's also not listed in `package.json`. Unintentional auto-import?

src/gui/search.ts
```diff
@@ -9,7 +9,7 @@
```
Jartreg, line 9: **Regarding the whole file:** This file doesn't seem to be imported anywhere, since there is also an identical file in `src/doc/`.

src/index.ts
```diff
@@ -0,0 +1,46 @@
...
+		}
+		ide.interpreter.run();
+	}else{
+		switch(window.location.search.slice(1)){
+			case 'doc':
+				doc.create(container);
+				break;
+			case 'unittest':
+				runTests(container);
+				break;
```
Jartreg, line 39: I'm not sure including the unit tests in `index.html` is a good idea. This currently adds 218 kilobytes to the bundle, which will be downloaded every time the IDE is used. This seems pretty unnecessary, since the unit tests are rarely run by users (have they ever been?).

Of course this could be solved by adding yet another bundle, but maybe it's not necessary at all to have users run unit tests, since we now have an automated test setup. I think it's more likely that users report an issue directly in the issue tracker before they find the instruction in the readme to run unit tests.

README.md
```diff
@@ -5,7 +5,7 @@ programming language.
 ## Getting Started
 TScript comes as a single html file. It does not require installation.
 Simply open the file in a modern browser and you are ready to go.
-[Click here for a quick test.](https://tglas.github.io/tscript/distribution/index.html)
+[Click here for a quick test.](https://tglas.github.io/tscript/)
```
Jartreg, line 8: This also needs to be reverted

Jartreg:
Some more things regarding `README.md`:

* [The links to the examples](https://github.com/TheHllm/tscript/blob/cfb3fdafc10fa0f4cfe0bfb3171b79fbf0dccfab/README.md#L25-L27) will no longer work. Should we link to the source files instead then?
* [The link to the unit tests](https://github.com/TheHllm/tscript/blob/cfb3fdafc10fa0f4cfe0bfb3171b79fbf0dccfab/README.md#L36) will be broken as well. However, see [my comment](https://github.com/TGlas/tscript/pull/57#discussion_r565937448) on those first.

src/index.ts
TheHllm: I would agree no one runs those. I am not even so sure if browser tests are needed anyway, it seems unlikely, that any of the tests pass in the mocha but not on a browser.

README.md
```diff
@@ -21,21 +21,18 @@ In TScript, "hello world" is a one-liner:
...
-directory. Demos of the examples:
- - [turtle graphics: snowflake](https://tglas.github.io/tscript/examples/demos/snowflake.html)
- - [canvas graphics: game of life](https://tglas.github.io/tscript/examples/demos/gameoflife.html)
- - [canvas graphics: 3D cube](https://tglas.github.io/tscript/examples/demos/cube3D.html)
-
+directory.
+ - [turtle graphics: snowflake](https://raw.githubusercontent.com/TGlas/tscript/master/examples/snowflake.tscript)
+ - [canvas graphics: game of life](https://raw.githubusercontent.com/TGlas/tscript/master/examples/gameoflife.tscript)
+ - [canvas graphics: 3D cube](https://raw.githubusercontent.com/TGlas/tscript/master/examples/cube3D.tscript)
+- [audio sime wave](https://raw.githubusercontent.com/TGlas/tscript/master/examples/audio.tscript)
```
Jartreg, line 28: Typo 😈 
```suggestion
- [audio sine wave](https://raw.githubusercontent.com/TGlas/tscript/master/examples/audio.tscript)
```

src/index.ts
```diff
@@ -1,4 +1,3 @@
-import * as _ from 'lodash';
 import { ide } from "./gui/ide";
 import { doc } from "./doc/doc";
 import { runTests } from "./tests/runtests_gui";
```
Jartreg, line 3: This would need to be removed as well
```suggestion
```

src/index.ts
TGlas: Browser test are not important to me. Testing in node should be fine. Users do not need the tests, and they should not be included in `index.html` and also not in `tscript.js`.

TGlas:
> Some more things regarding `README.md`:
> 
>     * [The links to the examples](https://github.com/TheHllm/tscript/blob/cfb3fdafc10fa0f4cfe0bfb3171b79fbf0dccfab/README.md#L25-L27) will no longer work. Should we link to the source files instead then?
> 
>     * [The link to the unit tests](https://github.com/TheHllm/tscript/blob/cfb3fdafc10fa0f4cfe0bfb3171b79fbf0dccfab/README.md#L36) will be broken as well. However, see [my comment](https://github.com/TGlas/tscript/pull/57#discussion_r565937448) on those first.

Not having the examples would be a pity, but I'd survive :) I agree that we don't need the unit tests.

TheHllm:
The only why i removed the examples is, that they would need to be compiled or bundled. I could however imagine having a way of letting the ide load a file via http by setting a get parameter. There should be no harm as long as the scripts don't run on their own this could also be useful elsewhere.

TGlas:
> The only why i removed the examples is, that they would need to be compiled or bundled. I could however imagine having a way of letting the ide load a file via http by setting a get parameter. There should be no harm as long as the scripts don't run on their own this could also be useful elsewhere.

I like that idea. It also automatically keeps the examples up to date without building them anew with each push. Of course, it sounds a bit dangerous at first to run code from a URL, but hey, TScript is a sandbox inside of a sandbox. Also, you could display a big "run" button with the URL so the user is in control. Alternatively, you could just open the IDE with the code in the editor and the right panels open and let the user run it as usual.

src/gui/standalone.ts
```diff
@@ -0,0 +1,27 @@
...
+    cv.parentNode.removeChild(cv);
+    document.body.innerHTML = "";
+    document.body.appendChild(cv);
+    cv.style.width = "100vw";
+    cv.style.height = "100vh";
+    cv.style.top = "0px";
+    setTimeout( ()=>{
+        cv.width = 250;
+        cv.height = 250;
+    }, 100);
```
Jartreg, line 26: This doesn't seem to be ported correctly. In exports, the canvas gets a size of 0x0...

Canvas Export

This is how it was done previously:

https://github.com/TGlas/tscript/blob/88af6a8e64703f7c25120f49406bf5a7bc4be62c/examples/demos/cube3D.html#L10645-L10646

Jartreg (changes requested):
Some more findings...

src/gui/standalone.ts
TheHllm: On Chrome Version 88.0.4324.104 x64 the previous way dose not work.  It will set the canvas to have no size. This is also the case for this repo. So i simply changed it to use a 250^2 canvas. I dont see how my code could lead to the canvas havin no size.

As to why the size needs to be adjusted, the code of the ide will adjust the size of the canvas if the container changes size, this happens when the body is cleared.

I dont see how i could fix that without chaining a lot and it seems that this behavior is browser dependent....
Seeing as fixing this is not the goal of this PR i would suggest adding a new issue etc.

src/css/ide.css
```diff
@@ -10,7 +10,6 @@ body {
 	padding: 0;
 	border: 0;
 	box-sizing: border-box;
-	overflow: hidden;
```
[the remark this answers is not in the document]
TheHllm: The hidden overflow hid the unitest. And as it did not change the appearance of the ide i removed it.

src/gui/standalone.ts
Jartreg: Part of the problem might be that you are using e.h. `cv.height` to set the height of the _canvas_. However, `cv` refers to the containing `div` element (see line 16). To refer to the canvas, you'd need to use `ide.canvas`.

TGlas: Badly named variable, sorry for that. Maybe rename into canvasContainer.

src/gui/ide.ts
```diff
@@ -0,0 +1,2670 @@
...
+				module.addMessage(err.type, err.type + " in line " + err.line + ": " + err.message, err.line, err.ch, err.href);
+			}
+		}
+	
+		if (program)
+		{
+			module.interpreter = new Interpreter(program, defaultService);
+			module.interpreter.service.documentation_mode = false;
+			module.interpreter.localStorage = window.localStorage;
+			module.interpreter.service.audioContext = new AudioContext();
```
Jartreg, line 757:
```suggestion
```
This can be safely removed. `audioContext` is already set in `defaultService.ts` ([here](https://github.com/TheHllm/tscript/blob/7ce9751006b088607458d5ede32757d34d6aa7f8/src/interpreter/defaultService.ts#L266)). Also, accessing `AudioContext` breaks the IDE in Safari.

src/index.html
```diff
@@ -0,0 +1,11 @@
+<html>
+    <!DOCTYPE html>
```
Jartreg, line 2:
```suggestion
<!DOCTPE html>
<html>
```

The doctype has to be in the beginning of the file. [see here](https://developer.mozilla.org/en-US/docs/Web/HTML/Quirks_Mode_and_Standards_Mode#how_does_mozilla_determine_which_mode_to_use.3f)

Jartreg (changes requested):

src/gui/standalone.ts
Jartreg: I just tested old exports and I can confirm: we moved from "it just works in firefox" to "it works everywhere _except_ for firefox"... So yes, this should get an issue after merging this 🤷.

src/tests/tests.ts
```diff
@@ -1492,7 +1494,7 @@
 			wait(100);
 			var t2 = time();
 			var d = t2 - t1;
-			if d < 90 or d > 130 then print("wrong");
+			if d < 50 or d > 200 then print("wrong");
```
Jartreg, line 1497: Is there a specific reason that you changed this? Is the timing less reliable when running in node?

TheHllm: Yes and now. Tests are run async. This speeds up execution without needing to change `Interpreter.chunk`.  That means many more tests are run at once. This also means that depending on scheduling the timing might be off.

TGlas:
Thanks a lot for the many changes and for the thorough review. It looks good to me. Do you both think that the PR is ready for the merge?

Jartreg:
Looks good to me as well!

I do have some suggestions regarding the directory structure, but I thought it's better to do that in a separate PR.

What I think what's left to do:
1. To fix the merge conflicts, @TheHllm needs to merge `master` into this again and include recent changes (#59, #61, anything else?)
2. We need to remember that just merging this will break the site until GH Actions have run and @TGlas has changed the branch for GH Pages. Maybe a way to solve this:
	1. Create a branch `gh-pages` from the current `master`
	2. Switch GH Pages to that branch
	3. After merging GH Actions will automatically overwrite it
	
	&rarr; But what do we do if something breaks?

TheHllm:
Changes are merged.
I don't know if creating a new branch etc. is needed. at most ci/cd will take 2 minutes. Changing the settings of the repo will take ~10s. 
Everything should work just fine, as the actions are run on githubs servers. And if anything goes wrong the commit can be reverted fast enough.
And github has some sort of cache in front of the github pages so that should also buy some time.

src/css/codemirror.css
```diff
@@ -393,3 +393,8 @@ span.CodeMirror-selectedtext { background: none; }
   font-size: 90%;
   padding: 1px 3px;
 }
+
+.CodeMirror-dialog button {
+  font-size: 90%;
+  padding: 1px 3px;
+}
```
Jartreg, line 400: This rule is duplicated (look above)

```suggestion

```

Jartreg (changes requested):
One little mistake while merging...

Jartreg:
@TheHllm Creating the branch on GitHub would also take just ≈10s, but I agree that in the best case everything should go fine.

Jartreg (approved):

Jartreg:
Since by now we have 60 commits here, with many of them being merges or related to very small fixes, would it be a good idea to do a squash merge?

TGlas:
Thanks for taking care of the last steps. It seems that everything is ready. However, before I run the merge, there is one thing I'd like to understand.

> I don't know if creating a new branch etc. is needed. at most ci/cd will take 2 minutes. Changing the settings of the repo will take ~10s.
> Everything should work just fine, as the actions are run on githubs servers. And if anything goes wrong the commit can be reverted fast enough.
> And github has some sort of cache in front of the github pages so that should also buy some time.

I understand that github-pages may be out of sync for a short time, but that does not bother me much. In the worst case it will show the old version, which is close enough. But which settings do I need to change after the merge? I guess that this is related to setting up an action for ci/cd -- or is this already specified in the PR with some magic and you refer to something else? I'd like to be certain to understand what I have to do in order to complete the process properly.

Jartreg:
The CI/CD setup should work automatically. It should only be necessary to change the branch for GitHub Pages after CD has completed (Settings → Options → GitHub Pages → Source).

(It would be possible to have a short meeting to talk about it...)

But one thing that you might have misunderstood: until the branch has been changed, the site will **not be accessible**.

TheHllm:
Github actions are defined in the .github folder. They will be run automatically on certain events (CI/CD is run on a push to master).
After they are done you only have to change the github pages.

Jartreg:
The individual steps would be:

1. (Squash) Merge
2. Wait for CD to finish
3. Change the branch in settings

Or:

1. Create a branch `gh-pages`
2. Change the branch in settings
3. (Squash) Merge

TGlas:
Thanks for the explanation. That sounds safe enough, I'll go ahead now.

Jartreg:
Aaaaaand it's live!! 🎉

TheHllm:
Is it just me or is github pages broken?

TGlas:
The "quick test" link in the readme does not work. Add a slash then it does.

TGlas:
Solved.
Once more, thanks a lot for the contribution!!

PR #61, opened 2021-01-30
Fix parse error for this-keyword in default parameters

### LINKED ISSUE #60: Internal Parser Error for default this-parameter
Creating a method with a parameter that is assigned the default value of `this` will result in an internal parser error.

Running this minimal example
```javascript
class Test {
  public:
    function test(argument = this) {
      return 4;
    }
}
```
will result in the following error:
```
error in line undefined: internal parser error; Cannot read property 'petype' of null
```

The simplest solution to this problem would probably be to disallow `this` as a default parameter.

Jartreg:
This fixes #60.

The reason for this bug is that `parse_expression` searches for a function, but doesn't expect it to be `null` ([here](https://github.com/TGlas/tscript/blob/88af6a8e64703f7c25120f49406bf5a7bc4be62c/source/tscript.js#L3794)). To fix this, `parse_function` and `parse_constructor` should pass the function itself as the parent when parsing default parameters.

This results in the expected error:

```
error in line 3: error in parameter declaration; default value must be a constant
```

### DIFF
--- a/distribution/index.html
+++ b/distribution/index.html
@@ -5295,7 +5295,7 @@ function parse_function(state, parent, petype)
 		if (token.type === "operator" && token.value === '=')
 		{
 			module.get_token(state);
-			let defaultvalue = parse_expression(state, parent);
+			let defaultvalue = parse_expression(state, func);
 			if (defaultvalue.petype !== "constant") state.error("/syntax/se-38");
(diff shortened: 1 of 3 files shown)


TGlas:
Thanks for the quick fix!

PR #63, opened 2021-02-02
Improve GitHub Action Workflows

Jartreg:
This...

* adds more descriptive labels to each step
* groups deployment commands in CD into _one_ step
* adds caching for NPM packages
* makes deployment commits show the actual github actions bot as the committer:
  [example commit](https://github.com/Jartreg/tscript/commit/0ad21c12ce922c4dfd343d81275995f9a7242854)
* removes the unnecessary matrix strategy from CD to prevent possible problems in the future. It's also not used in CI but I left it there to show that it's possible.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -20,6 +20,13 @@ jobs:
       uses: actions/setup-node@v1
       with:
         node-version: 15.x
+    - name: Set up NPM cache
+      uses: actions/cache@v2
+      with:
+        path: ~/.npm
+        key: npm-${{ hashFiles('**/package-lock.json') }}
+        restore-keys: |
+          npm-
     - name: Install dependencies
       run: npm ci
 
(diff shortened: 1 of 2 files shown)


TGlas:
I am not experienced with configuring github actions, therefore I do not feel confident to review this PR.
@TheHllm the original version of the files was yours. Could you have a look?

Jartreg:
To see it in action:

* [Successful CI run](https://github.com/TGlas/tscript/pull/63/checks)
* [Successful CD run](https://github.com/Jartreg/tscript/runs/1809236832?check_suite_focus=true)

TheHllm:
I don't know much about actions either, but everything seems to be in order.
It would be nice if we could somehow reuse the ci steps in the cd pipeline but i could not and still can't find a way to do that.

Jartreg:
Yes, unfortunately this doesn’t seem to be possible (in an easy way) 🤷.

PR #64, opened 2021-02-02
Improve webpack's settings for development

Jartreg:
Currently, the `dev`-script is supposed to watch in the background and rebuild when files change. However, the inlining plugins somehow don't re-emit their output, making the whole setup pointless. This is fixed by not using these plugins during development and emitting a separate javascript file instead.

This also enables us to use source maps during development! I chose the setting `eval-cheap-module-source-map`, because it shows the original source while supposedly being fast during rebuild ([see here](https://webpack.js.org/configuration/devtool/)), but that's still up for discussion.

<details>
<summary>Result in devtools</summary>

Firefox devtools showing the original source

</details>


Jartreg:
CC @TheHllm

PR #66, opened 2021-02-03
Fix undefined `this.assert`

### LINKED ISSUE #65: [BUG] Can't step into method calls
Example code:

```js
[].size();
```

When stepping through this, the interpreter get's stuck at "access of member size" and never reaches "function call". In the browser console, this error shows:

> TypeError: this.assert is not a function.
> (In `this.assert("type"===t.value.b.func.parent.petype,"[previewValue] invalid method object")`, `this.assert` is undefined)

Jartreg:
fixes #65

In `tscript.ts`, there are a some calls to `this.assert`, which is never defined:

https://github.com/TGlas/tscript/blob/adbaf4648bf33f77500e1aa5de97b5f00c28f7e2/src/tscript.ts#L475-L479

This has caused some errors while generating the stack view when stepping into function calls. Looking hat the history of these calls, it can be seen that they should refer to `ErrorHelper.assert` instead.

### DIFF
--- a/src/tscript.ts
+++ b/src/tscript.ts
@@ -1,5 +1,6 @@
-import { Typeid } from './helpers/typeIds';
+import { ErrorHelper } from "./errors/ErrorHelper";
 import { recApply } from './helpers/recApply';
+import { Typeid } from './helpers/typeIds';
 import { Interpreter } from './interpreter/interpreter';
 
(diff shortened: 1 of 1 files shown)


TGlas:
Thanks for the quick fix!

PR #71, opened 2021-02-06
Fix turtle and canvas reset issues

### LINKED ISSUE #70: turtle and canvas state are not reset of program start
The turtle position and orientation is not reset to the origin on program start. To reproduce the problem, simply run

    turtle.move(40);

multiple times in a row. Also, the line height is changed when setting the font:

    canvas.setFont("Helvetica", 50);
    canvas.text(50, 50, "bla\nbla");   # two lines to test line height

Running the program, removing the first statement and running it again demonstrates the problem.

Jartreg:
The turtle service is not properly reset in `clear()`, because `reset()` was
called after the interpreter was set to `null`.
Resetting both canvas and turtle directly after creating the interpreter
ensures that they are in a correct state.

This fixes #70.

### DIFF
--- a/src/ide/ide.ts
+++ b/src/ide/ide.ts
@@ -709,7 +709,6 @@ export let ide = (function() {
 			ctx.setTransform(1, 0, 0, 1, 0, 0);
 			ctx.clearRect(0, 0, module.turtle.width, module.turtle.height);
 		}
-		if (module.interpreter) module.interpreter.service.turtle.reset.call(module.interpreter, 0, 0, 0, true);
 	
 		let ctx = module.canvas.getContext("2d");
 		ctx.setTransform(1, 0, 0, 1, 0, 0);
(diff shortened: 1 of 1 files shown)


TGlas (commented):
Wouldn't it be better to give each interpreter instance its independent instance of services, with independent state? Currently there is no concrete need for parallel interpreter instances in the IDE, but combining a global service state with potentially multiple interpreters sounds dangerous. I can think of situations where this actually matters: at some point I created a system for emulating an online exam client-side in the browser, acting as a mock exam system (it is old and would require a major overhaul). In there I had an interpreter instance for the IDE combined with an independent interpreter instance for the automatic evaluation.

Therefore I propose to replace the static defaultService object with a createDefaultServices() function, which can be called whereever defaultService is used now. What do you think?

Jartreg:
I totally agree! With this I tried to add as few changes as possible, but it wouldn't have been a permanent solution.

TGlas:
I'll go ahead with that solution.

Jartreg:
If you don't mind I would like to propose a solution myself, because it gives us the oportunity to rewrite the services as Typescript classes.

TGlas:
On 06.02.21 12:14, Jartreg wrote:
> If you don't mind I would like to propose a solution myself, because it 
> gives us the oportunity to rewrite the services as Typescript classes.

Sure! I did only the quick fix. Just create an independent PR.

PR #73, opened 2021-02-06
Add missing CodeMirror Addons

### LINKED ISSUE #69: several hotkeys stopped working
At least in Firefox on Linux, hotkeys like Ctrl+D (toggle comment) and Ctrl+F (search) do not work any longer. In contrast, debugger-related hotkeys like F7, F8, and Shift+Ctrl+F11 work as expected.

Jartreg:
This adds back some CodeMirror addons that seem to have been overlooked in #57. There are three commands that I needed to change ([list of supported commands](https://codemirror.net/doc/manual.html#commands)):

https://github.com/TGlas/tscript/blob/06c45532b52748afbc75b3b71effd23f2311866a/src/ide/ide.ts#L2232-L2234

`unindent` is now simply `indentLess`, but I couldn't find anything on the other two. After trying what these shortcuts did in the ide before #57, they seem to be `goDocStart` and `goDocEnd`.

Also, all files are now importing CodeMirror with correct typings (without casting it to `any`).

fixes #69

### DIFF
--- a/src/ide/codemirror-tscriptmode.ts
+++ b/src/ide/codemirror-tscriptmode.ts
@@ -1,6 +1,5 @@
-import { default as cm } from "codemirror";
+import CodeMirror from "codemirror";
 
-let CodeMirror: any = cm;
 export const cmtsmode = CodeMirror.defineMode("text/tscript", function(config, parserConfig)
 {
 	const keywords = {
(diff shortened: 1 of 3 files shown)


TGlas:
Thanks!!

PR #75, opened 2021-02-13
GUI Improvements and Refactoring

manuel-fischer:
Here is some refactoring work and some changes to the user interface.
- The creation of dialogs/modals is moved to `tgui.ts`. 
- The positioning of the modals has been adjusted at the pixel-level, (✱)
- A dialog/modal now catches tab focus, pressing tab reaches the first element in the dialog again (✱)
- Capitalized first letter of all button texts
- Buttonbars with default buttons have been added, these provide a simple way to add buttons at the bottom of a dialog. They look like this (Here I replaced the default confirm dialog (Which has the buttons "OK" and "Cancel") with a custom one)

  grafik
  _Here the default button is "Yes" but might "No" be safer/a better option? What about naming the buttons "Discard" and "Cancel"_ (✱✱)

  grafik

Note that I have been working on those changes before the file structure of the repository changed. Those changes were originally in the improvements branch.

**(✱) this needs testing on other browsers!**
**(✱✱) this needs further investigation!**

### DIFF
--- a/src/ide/tgui.ts
+++ b/src/ide/tgui.ts
@@ -1222,14 +1222,14 @@ export let tgui = (function() {
 	// - buttons:           list of strings like ["Okay", "Cancel"], if this is not given, there is no button bar
 	//                      at the bottom.
 	//                      If the text of a button is not a fixed word, a name-text pair should be used
-	//                      [["Open", "Open File"], "Cancel"], the first is passed to onChoice and 
+	//                      [["Open", "Open File"], "Cancel"], the first is passed to onButton and 
 	//                      might be referred to in the default_button attribute.
 	//                      The second is only used as the visible button text.
 	//                      Note that an object/a dictionary is not used, because the order of the elements
 	//                      is important.
 	// - default_button     The name of the default button in buttons. It gets highlighted in blue.
 	// - enter_confirms     Boolean flag, true if pressing [Enter] should behave like clicking on the default button
-	// - onChoice           Handler function, that takes exactly one argument: the name of the button
+	// - onButton           Handler function, that takes exactly one argument: the name of the button
 	//                      on the buttonbar that has been pressed.
 	//                      It is also called if the close-button on the titlebar is clicked or [Escape]
 	//                      has been pressed.
@@ -1252,27 +1252,18 @@ export let tgui = (function() {
 		// dialog  -- the DOM object of the dialog contents
 		let control = Object.assign({}, description);
 		// handle default fields
-		if(!control.hasOwnProperty("onChoice")) control.onChoice = function(button) {};
+		if(!control.hasOwnProperty("onButton")) control.onButton = function(button) {};
 
 		let handleChoice = function(event, button)
 		{	
-			// by closing the current dialog before calling onChoice,
-			// onChoice itself is able to open additional dialogs in place
-			// of the current one
-			// if the onChoice returns true, the dialog is reopened, note
-			// that the content keeps being the same, stopModal and addModal
-			// do not invalidate the control object, thus the control object
-			// can simply be reused
-			// TODO: move stopModal to the bottom and call it conditionally
-			// with this dialog
-			tgui.stopModal(); // removes current dialog
-			let ret = control.onChoice(button);
+			// if the onButton returns true, the dialog is kept opened
+			let ret = control.onButton(button);
 			if(event)
 			{
 				event.preventDefault();
 				event.stopPropagation();
 			}
-			if(ret) tgui.startModal(control); // reopen current dialog
+			if(!ret) tgui.stopModal(control); // close current dialog
 			return false;
 		}
 		
@@ -1460,7 +1451,7 @@ export let tgui = (function() {
 		}
 	}
 
-	// Properties of description: prompt, [buttons], [default_button], title, onChoice...
+	// Properties of description: prompt, [buttons], [default_button], title, onButton...
 	// See `createModal` for more information about these properties
 	module.msgBox = function(description)
 	{
@@ -1471,6 +1462,7 @@ export let tgui = (function() {
 			"scalesize":      [0.20, 0.15], 
 			"minsize":        [300, 150],
 		}, description));
+
 		tgui.createElement({
 			"parent": dlg.content,
 			"type": "div",
@@ -1529,20 +1521,38 @@ export let tgui = (function() {
 			element.content.focus();
 	}
 
-	// Discard the topmost modal dialog.
-	// TODO: add parameter that identifies the modal to close,
-	//       which would allow to close dialogs inbetween, useful for the
-	//       buttonbars
-	module.stopModal = function()
+	// Discard a modal dialog.
+	// dialog -- Optional parameter of the dialog to close.
+	//           The default is the topmost dialog.
+	module.stopModal = function(dialog?)
 	{
-		if (modal.length == 0) throw "[tgui.stopModel] no modal dialog to close";
+		if(modal.length == 0) throw "[tgui.stopModal] no modal dialog to close";
 
-		// remove the topmost modal element
-		let element = modal.pop();
-		document.body.removeChild(element.dom);
-		
-		// restore previous active element
-		if(element.prevActiveElement !== null) element.prevActiveElement.focus();
+		if(dialog === undefined || dialog === modal[modal.length - 1])
+		{
+			// remove the topmost modal element
+			let element = modal.pop();
+			document.body.removeChild(element.dom);
+			
+			// restore previous active element
+			if(element.prevActiveElement !== null) element.prevActiveElement.focus();
+		}
+		else
+		{
+			let index = modal.findIndex(e => e === dialog);
+			if(index == -1) throw "[tgui.stopModal] not a modal dialog";
+
+			// The previous active element of the dialog above the current one
+			// is usually pointing to an element in this dialog to be removed.
+			// The next dialog inherits the previous active element. Such that,
+			// if all the dialogs would have been closed normally, the same active
+			// element gets focused, when the dialogs are closed out of order.
+			modal[index+1].prevActiveElement = modal[index].prevActiveElement;
+
+			// remove the modal element
+			modal.splice(index, 1);
+			document.body.removeChild(dialog.dom);
+		}
 
 		// remove the separator after the last modal dialog was closed
 		if (modal.length == 0) document.body.removeChild(separator);
--- a/src/ide/tgui.ts
+++ b/src/ide/tgui.ts
@@ -1211,7 +1211,8 @@ export let tgui = (function() {
 	// TODO: merge this with arrangePanels, that is also installed as a "resize"-callback
 	window.addEventListener("resize", centerAllModalDialogs);
 
-
+	module.modalClose = "tgui.modalClose";
+	module.modalHelp  = "tgui.modalHelp";
 	// Create a modal dialog. Similar to createPanel
 	// The description object has the following fields:
 	// - title:             text in the title bar
@@ -1219,6 +1220,7 @@ export let tgui = (function() {
 	// - minsize:           [width, height] minimum size of the dialog, 
 	//                      when the whole viewport is smaller, the viewport size is used
 	// - contentstyle:      object to add/override some styles to/of the content element
+	// - helpEnabled:       enables help button in the titlebar, TODO: also open Help with [F1]
 	// - buttons:           list of [name, text] pairs like [["delete", "Delete File"], ["cancel", "Cancel"]],
 	//                      if this is not given, there is no button bar
 	//                      at the bottom. The first element, the name, is passed to onButton and
@@ -1233,7 +1235,10 @@ export let tgui = (function() {
 	//                      on the buttonbar that has been pressed.
 	//                      It is also called if the close-button on the titlebar is clicked or [Escape]
 	//                      has been pressed.
-	//                      In this case the string "defaultClose" is passed as a parameter.
+	//                      In this case the string "tgui.modalClose" is passed as a parameter.
+	//                      If the help button is enabled, then "tgui.modalHelp" is passed as a parameter, whenever
+	//                      [F1] or the help button was pressed.
+	//                      
 	//                      Return true if the dialog should be kept opened, otherwise return false or nothing at all.
 	//                      
 	//
@@ -1254,7 +1259,7 @@ export let tgui = (function() {
 		// handle default fields
 		if(!control.hasOwnProperty("onButton")) control.onButton = function(button) {};
 
-		let handleChoice = function(event, button)
+		let handleButton = function(event, button)
 		{	
 			// if the onButton returns true, the dialog is kept opened
 			let ret = control.onButton(button);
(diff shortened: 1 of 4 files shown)


Jartreg:
Browser support on macOS:
✓ Works in Chrome, Firefox and Safari

src/ide/tgui.ts
```diff
@@ -1146,14 +1164,329 @@ export let tgui = (function() {
...
+				var focusable = Array(...control.content.querySelectorAll(focus_query));
+				if(control.hasOwnProperty("div_buttons"))
+				{
+					focusable.push(...control.div_buttons.querySelectorAll(focus_query));
+				}			
+				
+				if(focusable.length === 0)
+				{
+					(document.activeElement as any)?.blur();
+				}
```
Jartreg, line 1273: Are you sure `.blur()` is the right thing to do here? When there are no focusable elements in the dialog (which is already very unlikely), why not focus the container?

Jartreg (commented):
There are some things I'm not sure about:

When tabbing through a dialog, should you be able to focus the close button in the upper right? Currently that's not possible.

Centering an element is _really really_ something that should be done with CSS, but I think this would need a little change in the DOM structure.

Regarding the current restructurings:
Since you are essentially rewriting the dialog logic, you could have used the opportunity to move it to a separate file. Also, because you are adding new functions, they should have properly typed parameters (depending on how complex they would be). However, I think major restructurings to improve typings and split it up into smaller files are inevitable, so it might be better to do that later 🤷...

What do you think?

TheHllm:
I am for labeling the buttons with what they do.
VsCode dose this as well:
image

src/ide/tgui.ts
```diff
@@ -1146,14 +1164,332 @@ export let tgui = (function() {
...
+			{
+				let buttonName: string, buttonText: string;
+				if(typeof button == "string")
+					buttonName = buttonText = button;
+				else
+					[buttonName, buttonText] = button as [string, string];
+
+				let event_handler = control.onClose;
+				if(control.hasOwnProperty("on"+buttonName))
+					event_handler = control["on"+buttonName];
```
Jartreg, line 1328: When using TypeScript, this should **definitely** be avoided.

Jartreg (changes requested):
About how button click handlers are implemented:

You are currently using dynamic property access for the handlers of modal buttons. This works for now, because `control` is `any`. However, once we finally get to write proper interfaces for everything, it might get us into trouble. Dynamic property access _can_ be well typed with TypeScript, but it won't be pretty and there are much nicer solutions. Some examples:

1. Make the `buttons` option an array of objects, which contain the handler as an optional property. This might look like this:
	```ts
	tgui.msgBox({
		// ...
		buttons: [
			{
				label: "Discard",
				isDefault: true,
				onClick: onConfirm
			},
			{ label: "Cancel" },
		]
	});
	```

2. Make modals have only _one_ handler that gets passed the button as a parameter, and ```null``` if it was closed with the close button. Example:
	```ts
	tgui.msgBox({
		// ...
		buttons: ["Discard", "Confirm"],
		onClose(button) {
			if (button === "Discard")
				onConfirm();
		},
	});
	```

manuel-fischer:
> conflicts
Hello again,

Sorry, I was busy the last weeks and didn't came to work on this. Just now I am working on the dialog code, the idea with the single event handler is a very good idea I did not thought about. I am going to implementing this. It is more descriptive than the way it is currently implemented. It is also good, because it does not depend on dynamically generated event handler names.

Greetings
Manuel

manuel-fischer:
Message Box Icons

grafik
grafik

src/ide/tgui.ts
```diff
@@ -1146,14 +1167,401 @@ export let tgui = (function() {
...
+	function roundToPhysicalPixel(virtual_px: number)
+	{
+		return (virtual_px*window.devicePixelRatio | 0)/window.devicePixelRatio;
+	}
+
+	// Center a dialog at the center of the window
+	// Dialogs are aligned at integral pixel coordinates to
+	// avoid pixel-jittering
+	function centerModalDialog(dlg)
+	{
```
Jartreg, line 1184: [Again](https://github.com/TGlas/tscript/pull/75#pullrequestreview-589987254), this should _definitely_ be done in CSS.
Something like this would work:

```css
.tgui-modal {
	  width: ...%;
	  height: ...%;
	  min-width: ...;
	  min-height: ...;
	  position: absolute;
	  top: 50%;
	  left: 50%;
	  tranform: translate(-50%, -50%);
}
```

However, there would be some issues when the screen is too small for the `min-`size to fit. To get this 100% right with CSS, we'd need to change the DOM structure. I think this is out of scope for this PR, so I don't want to block merging with this.

src/ide/tgui.ts
```diff
@@ -1146,14 +1167,401 @@ export let tgui = (function() {
...
+					else if(event.shiftKey && focusIndex === 0)
+					{
+						focusable[n-1].focus();
+						event.preventDefault();
+					}
+				}
+			}
+		});
+
+		control.handleClose = (event) => handleChoice(event, "defaultClose");
```
Jartreg, line 1314:
```suggestion
		control.handleClose = (event) => handleChoice(event, null);
```

Here, I would simply pass `null`, because _no button_ was chosen and `null` won't collide with button names.

Jartreg (changes requested):

src/ide/tgui.ts
TGlas: I don't think that a programming IDE needs to work on a device without a proper keyboard. Sure, that's not well defined. Anyway, assuming a minimal screen size of, say, 1280x800 (12 years old Macbook) seems safe to me. If you object, about 20 years back it was 1024x768. That's good enough for a dialog. No need to go down to 320x200, that was 28 years ([DooM](https://en.wikipedia.org/wiki/Doom_(1993_video_game))), and even 39 years ([C64](https://en.wikipedia.org/wiki/Commodore_64)) ago. You may be out of luck on a phone. Don't code on a phone :)

More seriously: If you have a good way of solving this then I'd appreciate, but I don't think that we should bother too much.

src/ide/tgui.ts
```diff
@@ -1146,14 +1167,401 @@ export let tgui = (function() {
...
+	//                      when the whole viewport is smaller, the viewport size is used
+	// - contentstyle:      object to add/override some styles to/of the content element
+	// - buttons:           list of [name, text] pairs like [["delete", "Delete File"], ["cancel", "Cancel"]],
+	//                      if this is not given, there is no button bar
+	//                      at the bottom. The first element, the name, is passed to onButton and
+	//                      might be referred to in the defaultButton attribute.
+	//                      The second is only used as the text displayed on the button.
+	//                      It might be a translated text.
+	//                      Note that an object/a dictionary is not used, because the order of the elements
+	//                      is important.
```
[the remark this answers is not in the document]
manuel-fischer: I do not think that this is a good idea, because the text on the button might be generated dynamically and the text itself might be a translation into another language like German. When doing this, the code calling `tgui.createModal`/`tgui.msgBox` is going to get more complex and the text in the `buttons` and in `onButton` might get out of sync.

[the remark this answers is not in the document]
Jartreg: I see your point, it's probably just that I don't like using tuples for this. I guess it's the right choice then for simplicity 👍.

src/ide/tgui.ts
Jartreg: That's exactly why I think this little CSS would be enough. Doing it in JS just to get it subpixel-perfect seems pretty overkill to me 😄

manuel-fischer:
I thought about using "defaultClose"/"tgui.modalClose" for when other buttons like the help button are added. But I think using `null` would be also a good idea, but I need to think about how it would work with the help button. When adding a property like `helpPath` there is the problem that the opening of the help/documentation is handled in `ide.ts`. When thinking about the dependency graph of the various files, `tgui.ts` should not have any dependencies on `ide.ts`, because `ide.ts` is built upon `tgui.ts`. A possible way would be to add a handler for the help. What do you think? How would you implement this? And is "tgui.modalClose" a better option than "defaultClose"?

src/ide/tgui.ts
manuel-fischer: I think if I wouldn't have worked on the CSS of the dialogs, I wouldn't have noticed pixel jittering, which appears when the browser size changes. I think no one would have ever noticed it :D. It would definitely be much cleaner code if the pixel perfect alignment was only done in CSS, but in CSS there is no possible way to do this alignment. In Firefox the pixel jittering is very noticeable, in Chrome and probably other browsers, it is not as visible. I also think it is a little overkill now. So should I discard this code and replace it back with CSS, similar to as it is in the main branch now? What do you think?

TGlas:
> I thought about using "defaultClose"/"tgui.modalClose" for when other buttons like the help button are added. But I think using `null` would be also a good idea, but I need to think about how it would work with the help button. When adding a property like `helpPath` there is the problem that the opening of the help/documentation is handled in `ide.ts`. When thinking about the dependency graph of the various files, `tgui.ts` should not have any dependencies on `ide.ts`, because `ide.ts` is built upon `tgui.ts`. A possible way would be to add a handler for the help. What do you think? How would you implement this? And is "tgui.modalClose" a better option than "defaultClose"?

I totally agree that the gui layer should not need to know about the ide. A handler solution sounds good to me. For "tgui.modalClose" vs. "defaultClose" I don't have a strong opinion.

Jartreg:
I still think `null` is best suited for the closing case. Since a modal can be closed in several ways, passing `null` best represents the fact that none of the intended buttons was pressed.

This different for the help button. Since it is a “magical” button, a magic string value works here. However, I agree that adding a special handler as an option instead is an even better solution. I’m not even sure whether the modal should be closed when pressing the help button.

src/ide/tgui.ts
Jartreg: Do you mean by "pixel jittering" that the dialog jumps around a little _while_ resizing the browser window? I think this is unfortunately normal behaviour and can't really be "fixed". For me it's even more noticeable in the JS version, since it doesn't run faster than the browser doing CSS layout.

If you mean that it's not positioned on pixel boundaries and causes anti-aliasing (does it even?), then you shouldn't worry about that at all, since it's highly dependent on the browser's rendering infrastructure. Trying to "fix" that will probably cause more problems in the long term than it will solve.

For what to do in CSS, you could either use my proposal from above, or stick with how it's currently done. I think both solutions should work fine...

manuel-fischer:
Now `buttons` is changed to an array of objects, this might be the cleanest solution, because no identifier-strings are passed around and every button object contains its own handler directly. For buttons that should simply close the dialog like "Cancel"/"Close", the handler should not be given (or a noop). Cleanup and other further work, that is always done while closing the dialog, like saving the configuration, would be handled in the `onClose` handler. Both `onClick` and `onClose` can cancel the closing of the dialog by assigning `true` to `dialog.keepOpen`. `onHelp` is its own independent handler.

src/ide/tgui.ts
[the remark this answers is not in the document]
manuel-fischer: I thought about it and came to the conclusion, that objects instead of tuples are more expressive and would resolve the problem of the button identifier strings, because every button could have its own handler now.

Jartreg:
> Both `onClick` and `onClose` can cancel the closing of the dialog by assigning `true` to `dialog.keepOpen`.

Why not do it with a `return true` in `onClose` and `onClick`? Is there a special reason that it has to be a property? I can see this as a possible source of bugs:

```ts
tgui.createModal({
	...everythingElse,
	buttons: [
		{
			text: "Confirm",
			onClick(dlg) {
				if (something) dlg.keepOpen = true;
			}
		},
		{
			text: "Cancel"
		}
	]
});
```
Looks completely fine, right?

Now imagine the following scenario:
* The user clicks "Confirm", but because of some condition, we want the dialog to stay open and assign `dlg.keepOpen = true`.
* The user decides otherwise and clicks "Cancel". There is no click handler for the "Cancel"-button, because there is nothing to do. However, because be assigned `true` to `keepOpen` before, the dialog will stay open for till the end of time.

manuel-fischer:
>     * The user decides otherwise and clicks "Cancel". There is no click handler for the "Cancel"-button, because there is nothing to do. However, because be assigned `true` to `keepOpen` before, the dialog will stay open for till the end of time.
`dialog.keepOpen` is set to `false` before any handler starts, so this would not be a problem.

Jartreg:
Oops, sorry I overlooked that.

src/ide/tgui.ts
manuel-fischer: Now I changed it to behave similar to when the dialog is opened. The `blur` is now not needed anymore, because the content `div` of the dialog is made explicitly focusable by Javascript with `tabindex="-1"`, in this case `.focus()` should not fail. Fix: The `event.preventDefault()` was missing. I have tested this by temporarily removing the buttons from the `[F1]` popup, that dialog would have no focusable elements.

manuel-fischer:
> There are some things I'm not sure about:
> 
> When tabbing through a dialog, should you be able to focus the close button in the upper right? Currently that's not possible.
> 
> Centering an element is _really really_ something that should be done with CSS, but I think this would need a little change in the DOM structure.
> 
> Regarding the current restructurings:
> Since you are essentially rewriting the dialog logic, you could have used the opportunity to move it to a separate file. Also, because you are adding new functions, they should have properly typed parameters (depending on how complex they would be). However, I think major restructurings to improve typings and split it up into smaller files are inevitable, so it might be better to do that later 🤷...
> 
> What do you think?

1. It is unusual that the [x] button and now maybe also the [?] button is focusable by [Tab]. For this the keys [Esc] and [F1] are commonly used instead.

2. I am unsure about the centering of the dialogs, that is now done in Javascript. Maybe it might be removed or refactored. Please let me know what would be the best option. Note that it was done with CSS previously.

3. I think that common refactoring like moving dialog logic into their own file might be now too much for this pull request. This should be done in its own pull request when other things might be restructured too.

Is there anything other that might be of discussion? Otherwise I might just make final adjustments and call this pull request as completely finished.

Jartreg:
> It is unusual that the [x] button and now maybe also the [?] button is focusable by <kbd>Tab</kbd>. For this the keys <kbd>Esc</kbd> and <kbd>F1</kbd> are commonly used instead.

I know it's unusual for window buttons managed by the operating system, but it's different on the web. Here, _everything_ should be accessible by keyboard. The <kbd>Esc</kbd> and <kbd>F1</kbd> keys are good as _additional_ shortcuts, but all buttons should still be able to be reached using the <kbd>Tab</kbd> key.
**However**, since I think it's mostly an advantage for people using screen readers and these people are probably lost on the TScript IDE anyway, so we might get away with not changing it (yet?). These kind of detailed accessibility related changes might be better suited for an overall accessibility overhaul.

> I am unsure about the centering of the dialogs, that is now done in Javascript. Maybe it might be removed or refactored. Please let me know what would be the best option. Note that it was done with CSS previously.

See https://github.com/TGlas/tscript/pull/75#discussion_r623636022. I'm still not sure what exact kind of problems you are talking about, but I'm sure that using JavaScript for this will most likely only cause trouble in the future. Centering an element is such a basic thing and CSS was made for that.

> I think that common refactoring like moving dialog logic into their own file might be now too much for this pull request. This should be done in its own pull request when other things might be restructured too.

You are right. There's still much to do to improve the structure of the project (you might've noticed that I like keeping files small :wink:) that should be done in separate pull requests.

src/ide/tgui.ts
```diff
@@ -1145,15 +1171,407 @@ export let tgui = (function() {
...
+		
+		dialog.addEventListener("keydown", function(event)
+		{
+			if(event.keyCode === 9) // TAB characterSet
+			{			
+				let focus_query = 'button, [href], input, select, textarea';
+				// collect all focusable elements inside the dialog, except titlebar buttons
+				var focusable = Array(...control.dom.querySelectorAll(focus_query));
+				focusable = focusable.filter(element => element.tabIndex !== -1);
+				console.log(focusable);
```
Jartreg, line 1300: You probably forgot to remove the `console.log` here:
```suggestion
```

Jartreg (changes requested):
Last little nitpick, everything else looks good to me!

Jartreg (approved):
:+1:

manuel-fischer:
I think, that then this pull request is complete, even if I have many other ideas like a dark mode for the IDE. Those would be implemented in another pull request.

TGlas:
Thanks, highly appreciated!

Issue #85, opened 2021-07-27
Misleading comment in example for se-35


Jartreg:
The documentation page for `se-35` currently has a misleading example in its example:

https://github.com/TGlas/tscript/blob/ea3ef0876ce2802ef95bbf00af43834cbfbfcb4f/src/doc/errors.ts#L597-L602

The comment says "behind index a comma or closing bracket is expected", but the actual error is the missing opening bracket.

This was introduced in 45ce189f66df11b7e3d87ba4adecca6a7901f6df (maybe a copy-paste mistake from [`se-31`](https://github.com/TGlas/tscript/blob/45ce189f66df11b7e3d87ba4adecca6a7901f6df/src/doc/errors.ts#L497)?).

PR #87, opened 2021-07-28
fixed merge conflicts and fixed error 84

duckmopsi:


Jartreg:
Is this a good idea? The merge conflicts are already resolved. Trying to merge these changes again will surely result in even more conflicts (as you can already see below). The fix for the example comment should rather be it's own PR, don't you think?

Jartreg:
Another question — by "error 84", so you mean issue #85?

PR #88, opened 2021-08-01
finally fixing the example error

duckmopsi:


src/doc/errors.ts
```diff
@@ -596,7 +596,7 @@ function example(index) {
             <tscript do-not-run>
 var array = [0, 1, 2, 3, 4, 5];
 function example(index) {
-    return function index] () {         # here the error occurs, because behind index a comma or closing bracket is expected
+    return function [index () {         # here the error occurs, because behind index a comma or closing bracket is expected
```
Jartreg, line 599: I think you are mistaking error IDs here. This is the documentation for error `se-35` (see line 585). Your new example code yields an error because of a missing comma or closing bracket, which is error `se-31`:

The new example code causing se-31

You can see that this is the error message for `se-31`:

https://github.com/TGlas/tscript/blob/9f3dee15b0dfcc9424ed00092b416fa4784c3b1f/src/lang/errors/ErrorHelper.ts#L152-L153

And the documentation for `se-31` is elsewhere:

https://github.com/TGlas/tscript/blob/9f3dee15b0dfcc9424ed00092b416fa4784c3b1f/src/doc/errors.ts#L535-L557

This means that the example code _was_ correct before, only the comment didn't fit:

The old example code causing se-35

https://github.com/TGlas/tscript/blob/9f3dee15b0dfcc9424ed00092b416fa4784c3b1f/src/lang/errors/ErrorHelper.ts#L158-L159

It's not the example code that needs to be changed but just the comment. Maybe something like this would work:

```suggestion
    return function index] () {         # here the error occurs, because before `index`, an opening bracket is expected
```

Jartreg (changes requested):

PR #93, opened 2021-09-11
Update NPM dependencies

Jartreg:
Now that all exams are over we can risk breaking things again! 😜

This updates dependencies in `package.json` and `package-lock.json`.
Some adjustments are necessary because TypeScript now complains that
caught exceptions are of type `unknown` by default.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,31 +5,53 @@
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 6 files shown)


TGlas:
Thanks!
Btw, I maintain a stable version on the webserver, so we can anyway break things any time :)

PR #95, opened 2021-09-19
Dark Theme for the IDE

manuel-fischer:
One thing that was missing from the TScript IDE is a dark theme, which is provided by most software/websites nowadays. It is considered more friendly for the eyes and various newer display technologies would consume less energy. I personally think that a dark theme is more visually pleasing.

So I have implemented a dark mode for the IDE and the documentation:

grafik

grafik

grafik

It is implemented by adding the `.dark-theme` class to the body tag, all css files are copied into the `src/ide/css-dark/` folder and changed in such a way that mostly colors are changed. Also various icons automatically adapt to the dark theme, for this I have reorganized all the icon code into `src/ide/icons.ts`.

What do you think about my implementation of the dark theme. Some code might need to be refactored.


TGlas:
I always find that argument funny. In the early 1980es, PCs used green or bernstein (monochrome) monitors with black background. When color displays came up (16 colors, later 256), the argument for black font on white background was that it is much friendlier to the eyes, resembling work on paper. The history put aside, it is a good idea to leave the decision to the user.

Regarding the implementation, I don't know what would be the best way to do this. One thing that comes to my mind is to change the canvas implementation of the toolbar buttons into data URLs of backgrounds. The advantage would be that they can be switched to dark with CSS means, which makes the implementation less immersive and more self-contained.

A detail: to make things future proof, how about changing the checkbox in the config dialog into a dropdown list?

src/ide/css-dark/ide.css
```diff
@@ -0,0 +1,128 @@
...
+
+body.dark-theme *.ide-program-current {
+	background: #653;
+}
+
+body.dark-theme *.ide-turtle {
+	background: #333;
+}
+
+.ide-tutorial {
```
Jartreg, line 62: Did you forget `body.dark-theme` here?

src/ide/css-dark/tutorial.css
```diff
@@ -0,0 +1,50 @@
+.tutorial-nav {
```
Jartreg, line 1: Does this file need to be here? Nothing seems to be changed...

src/ide/ide.ts
```diff
@@ -2160,27 +1931,39 @@ export let ide = (function () {
 				"text-overflow": "ellipsis clip",
 
 				"text-align": "center",
-				background: "#fff",
+				//background: "#fff",
 			},
 		});
+
+		module.programstate.setCss = function (cls) {
```
Jartreg, line 1938: Maybe call this `setCSSStateClass` or something else more descriptive? When I read `setCss` I'd think of setting styles directly...

src/ide/tgui.ts
```diff
@@ -130,6 +133,13 @@ export let tgui = (function () {
 			element.tabIndex = description.tabindex;
 		}
 
+		// html tag attributes, TODO: should these be in the description
+		// object directly?
+		if (description.hasOwnProperty("attributes")) {
+			for (let a in description.attributes)
+				element.setAttribute(a, description.attributes[a]);
+		}
+
```
Jartreg, line 142: Did you use this in your implementation of dark mode?

src/ide/tgui.ts
```diff
@@ -202,6 +212,27 @@ export let tgui = (function () {
...
+			let ctx = canvas.getContext("2d");
+			ctx.scale(zoom, zoom);
+		}
+	}
+
+	function renderCanvas(canvas, draw) {
+		if (typeof draw === "string") {
+			console.assert(draw.substring(0, 5) == "icon:");
+			let id = draw.substring(5);
+			icons[id](canvas, module.dark_theme);
```
Jartreg, line 232: What I dislike here is the use of strings to identify the icons. I think it would be better if you left the drawing functions in the description object and used a [`WeakMap`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap) to assign the drawing functions their canvas elements. When changing the theme you could then use that `WeakMap` again to get the drawing functions belonging to each canvas.

Another, more self-contained approach might be to emit a custom event on `body` when the theme changes with each canvas listening to it. This might be more complex however...

Jartreg (changes requested):
I'd guess a lot of people will like this 🎉

Regarding the styles in the dark mode CSS files: I don't think you need the universal selector which you added to each style block and it adds a lot of unnecessary noise to the style sheets. (Or did you test it without and needed it for specificity?)

Which means I would propose changes like this:
```diff
- body.dark-theme *.class {
+ body.dark-theme .class {
```

And I'm pretty sure that you definitely don't need it when selecting IDs:
```diff
- body.dark-theme *#id {
+ body.dark-theme #id {
```

I also support the argument against a checkbox in the settings dialog. A drop-down or radio buttons (which I would prefer) could be used to add the option to detect the theme automatically in the future. We'd also need to change the settings object in localStorage then...

src/ide/tgui.ts
manuel-fischer: I have used strings here, because those can be stored in the dom. I think about changing it in such a way, that strings might be used under the hood, but when assigning an icon, the elements in the icons module are accessed.

src/ide/tgui.ts
manuel-fischer: I have used it when I changed the layout of the check boxes. Before, the description object could not hold various html attributes, because those would simply be ignored in this function. So I added an explicit attributes tag, which contains html attributes that would definitely be applied to the element. Maybe it wouldn't be a bad idea to move them back in to the enclosing description object, but then it might be difficult to differenciate between custom description attributes and html attributes.

src/ide/tgui.ts
Jartreg: Again, you don't need to move the drawing functions to a new file and use strings. If you used a `WeakMap` to store the drawing functions, you could just keep them where they were and change much less.

#### Example
You create a `WeakMap` to store the functions:
```ts
type IconDrawingFunction = (canvas: HTMLCanvasElement, dark: boolean) => void;
const iconDrawingFns = new WeakMap<HTMLCanvasElement, IconDrawingFunction>();
```

You store the drawing function for the canvas:
```ts
iconDrawingFns.set(canvas, draw);
draw(canvas, isDarkTheme(module.theme));
```

And when the theme changes, you can just get them from the `WeakMap` again:
```ts
const draw = iconDrawingFns.get(canvas);
draw(canvas, isDarkTheme(newTheme));
```

Doing it this way has a much smaller impact and allows the icon drawing functions to stay where they are.

In addition, it might be better to keep them where they are. You can currently see whether and where they are used, because they are directly referenced and mostly even declared inline. When moving them away it gets a little harder to tell where a specific icon is used and you need to remember removing one once it's no longer used.

Jartreg: Again, you don't need to move the drawing functions to a new file and use strings. If you used a `WeakMap` to store the drawing functions, you could just keep them where they were and change much less.

#### Example
You create a `WeakMap` to store the functions:
```ts
type IconDrawingFunction = (canvas: HTMLCanvasElement, dark: boolean) => void;
const iconDrawingFns = new WeakMap<HTMLCanvasElement, IconDrawingFunction>();
```

You store the drawing function for the canvas:
```ts
iconDrawingFns.set(canvas, draw);
draw(canvas, isDarkTheme(module.theme));
```

And when the theme changes, you can just get them from the `WeakMap` again:
```ts
const draw = iconDrawingFns.get(canvas);
draw(canvas, isDarkTheme(newTheme));
```

Doing it this way has a much smaller impact and allows the icon drawing functions to stay where they are.

In addition, it might be better to keep them where they are. You can currently see whether and where they are used, because they are directly referenced and mostly even declared inline. When moving them away it gets a little harder to tell where a specific icon is used and you need to remember removing one once it's no longer used.

manuel-fischer:
> Regarding the implementation, I don't know what would be the best way to do this. One thing that comes to my mind is to change the canvas implementation of the toolbar buttons into data URLs of backgrounds. The advantage would be that they can be switched to dark with CSS means, which makes the implementation less immersive and more self-contained.

I suggest to use SVG images for the icons, those are scaleable and allow CSS styling. I am going to convert the canvas code to svg images.

TGlas:
SVG sounds like a good solution.

src/ide/tgui.ts
manuel-fischer: I suggest that the icons by themselves should be in a separate file, as I have currently implemented it here. This way the ide/gui source code is not cluttered as much. But I see the advantage of inline icons. Also when adding a new toolbar button, only one file needs to be changed.

Jartreg:
> I am going to convert the canvas code to svg images.

But does it need to be in this PR? The title of this PR says dark mode, so I would suggest you submit a working version with the changes concerning this topic and create a separate PR when you converted the icons. Doing too much unrelated stuff in one PR is IMHO not a good idea. If you're saying that you're going to convert the icons anyway than I'd be fine with your current handling of icons in the meantime 🤷🏼‍♂️.

manuel-fischer:
> But does it need to be in this PR? The title of this PR says dark mode, so I would suggest you submit a working version with the changes concerning this topic and create a separate PR when you converted the icons. Doing too much unrelated stuff in one PR is IMHO not a good idea. If you're saying that you're going to convert the icons anyway than I'd be fine with your current handling of icons in the meantime 🤷🏼‍♂️.

I can create an additional PR for the icons. The question is, should I move the icons back to where they were, or do you think they could be in their own file, `icons.ts`?

I am currently converting/testing the icons in my testing-repo https://github.com/manuel-fischer/svg-test and would create an additional PR where I implement those back into the ide.

src/doc/cheatsheet.ts
```diff
@@ -47,7 +47,7 @@ export const doc_cheatsheet: Documentation = {
 	</table>
 	<table class="nicetable">
 	<tr><th><a href="?doc#/library">standard library</a></th><th>topic</th></tr>
-	<tr><td><a href="?doc#/library/core">code functions</a></td><td>messages, time, errors, files, events</td></tr>
+	<tr><td><a href="?doc#/library/core">core functions</a></td><td>messages, time, errors, files, events</td></tr>
```
Jartreg, line 50: I'm honestly not a fan of unrelated drive-by changes, but since it doesn't seem to bother anybody else and is technically not wrong just keep it here 🤷🏼‍♂️

src/ide/index.html
```diff
@@ -5,7 +5,7 @@
 		<meta name="viewport" content="width=device-width, initial-scale=1" />
 		<title>TScript IDE</title>
 	</head>
-	<body>
+	<body style="overflow: hidden">
```
Jartreg, line 8: What is this for? Shouldn't it rather be in `ide.css`?

Jartreg (approved):

src/ide/index.html
manuel-fischer: This might also be a little unrelated, it removes the double scrollbars in the documentation, those only appear in Firefox. I think that it is a bug/rounding error in the CSS engine of Firefox. It might be suitable for another PR, what do you think?

Jartreg: If it's unrelated, I think it should be separated. But if nobody else complains just keep it here and do it differently in the future 🤷🏼‍♂️

Jartreg (approved):

manuel-fischer:
One thing that also changed is how the separators on the toolbar are placed, it now places the separators between the toolbar buttons, which have a different group.

Also the config button was added to the button list. It also got a new hotkey (control-p), which can now be changed, because it is now in the button list.

TGlas:
Thanks for another great contribution!

PR #96, opened 2021-09-27
SVG Icons & Dark Theme Detection

manuel-fischer:
SVG Icons are scaleable and can be styled by css. This way the code for updating icons is not needed. This cleans up various code related to icons in `tgui.ts`.

The icon drawing itself is similar to how it was before. I have implemented the `SVGDrawingContext`, that has functions like `rect` and `circle`, that have similar syntax to the canvas drawing context. Except that for example `rect` directly creates the `rect`-SVG primitive and there is no `fillRect`. The filling behavior is set by the style. Complex paths can be drawn with `beginPath`, `moveTo`, `lineTo`, `arc`..., but `endPath` is used instead of `stroke` or `fill`.

Please let me know how the implementation can be improved. Do you prefer to include direct SVG or programmed SVG as it currently is? What do you think about my `SVGDrawingContext` implementation?


Jartreg:
I somehow expected you'd convert the icons to actual SVG code...
Why don't you use the `SVGDrawingContext` to generate the SVG code and then include only that? I don't think we'd need this for anything else and it unnecessarily increases size and runtime costs. Am I missing something?

manuel-fischer:
> I somehow expected you'd convert the icons to actual SVG code... Why don't you use the `SVGDrawingContext` to generate the SVG code and then include only that? I don't think we'd need this for anything else and it unnecessarily increases size and runtime costs. Am I missing something?

Do you mean that the svg/xml code is created at build time from the current code that is in `icon.ts`?
I think that it wouldn't make a difference, but I am open to implement this.
The performance behavior would be as before, because the svg graphics are built only when the html element is created, similar how now the canvas icons are drawn. Also I think that the minified source code of the icons is at least as small as the corresponding svg/xml code.

The svg/xml code could be created in an `preval` block or similar, which sounds like an interesting adventure to explore.

Jartreg:
I meant you could use `SVGDrawingContext` to write a script that generates the `svg/xml` **now**, and commit **only that**. Nothing at build time.

If we were to go a step further, using some [✨ webpack reconfiguration ✨](https://webpack.js.org/guides/asset-modules/), you could even have your script generate `.svg`-files (e.g. `icons/dock-left.svg`) and import them from code. The benefit of this: SVG files look nicer in the file tree, as they can be previewed.

src/ide/svg.ts
```diff
@@ -0,0 +1,179 @@
...
+		if (properties.hasOwnProperty(p)) el.setAttribute(p, properties[p]);
+	}
+	return el;
+}
+
+export class SVGDrawingContext {
+	public readonly svg: SVGSVGElement;
+	#classname: string;
+	#style: string;
+	#path?: string;
```
Jartreg, line 23: Concerning ES private class member syntax: please don't use it. It's only a very recent feature (Firefox has it since July) and Safari doesn't yet support private class methods.

In exchange, just use TypeScript's [visibility modifiers](https://www.typescriptlang.org/docs/handbook/2/classes.html#member-visibility). They ensure the correct visibility at compile time which is good enough.

Jartreg (changes requested):
I'm perfectly fine with this so far!

One thing:
I don't think we need the generation code (`gen-icons.ts`, ...). The SVG code in `icons.ts` should be enough. Should we ever need to regenerate them (I don't think so), that's what we have git for.

TGlas:
> One thing:
I don't think we need the generation code (gen-icons.ts, ...). The SVG code in icons.ts should be enough. Should we ever need to regenerate them (I don't think so), that's what we have git for.``

That's true, but in this case I prefer a redundant solution. A more expectable scenario is that we want to add a new icon. Having the generate-script in the project is then convenient. Adding the SVG icons to git is also a good thing, since it simplifies the build process. Let's do both.

Jartreg:
Well I'd be fine with it for the current solution with everything being in `icons.ts` :+1:

However, if we were to move to actual SVG files in the future, I'd suggest having _them_ as the actual source of truth. Meaning: when adding an icon, you add SVG code, not code that generates it. Admittedly, this is currently not very practical.

Jartreg (approved):

manuel-fischer:
One advantage of using SVG files is that you can edit the SVG files with various SVG editing sofware. But using such Software might lead to a problem, because you might not have the complete control of the styling. Writing the SVG files explicitly could also lead to "code" duplication in the icons, eg. for "Save" and "Save As". In my opinion the SVG format by itself is a little problematic when using paths for example: you need to know what the various line commands mean. The drawing code in 'gen-icons.ts' can be better understood by people, that do not know the SVG format. And another advantage is that the generation code can be typechecked very easily, compared to checking the format/syntax of the SVG source.

Jartreg (changes requested):
Oops, something that I totally missed: the automatic detection of dark mode is totally unrelated to this (hint: you notice it if you write something like `missing from previous pull request` 😜). In addition to that, there are some  points that I'd like to talk about concerning your implementation. Question is: split it up into another PR to keep things separated or let this slip through again and talk about this here?

manuel-fischer:
I think that for such a small feature a new PR is not really needed, as in my opinion many different PR might reduce productivity. I have renamed the PR to cover the automatic detection of the dark theme, because both are related to the UI.

manuel-fischer:
## Rename: default → light; auto → default

It might be useful that only settings, that are changed or explicitly set by the user are stored in the "tscript.ide.config", this way implementation changes considering the settings are more flexible.

## Change-Event

When the theme would be updated when the system setting changes, you might also consider to update the theme when the settings are changed in another browser tab. But then you might also think about updating other settings as well. It might make it more complicated.

Jartreg:
> When the theme would be updated when the system setting changes, you might also consider to update the theme when the settings are changed in another browser tab.

No, I don't. I just want the theme to stay in sync with the system.

> But then you might also think about updating other settings as well.

No, I don't. I just want the theme to stay in sync with the system.

To be clear: I'm not talking about automatically changing the setting. The idea is that when `default` is selected, the theme adapts to the system, **and follows when it changes.** But I'd also be fine if you don't want to do it now. I, too, want this PR merged 😊.

PR #97, opened 2021-10-01
Move the IDE to the root directory of the deployment

Jartreg:
We had discussed before in #57 that it would be neat to have the IDE at `https://tglas.github.io/tscript` instead of `https://tglas.github.io/tscript/distribution`. Because we are in-between semesters, I think we can finally do the move.

Old bookmarks and URLs are still supported and redirected by `redirect.html`, which is moved to `distribution/index.html` during deployment. It redirects to the new URL **keeping all query and hash strings**.

I renamed the output directory from `distribution` to the more common `out`. We could also use `build` or something similar as you like. ~~(Attention: at first, git will notice old `distribution` directories because it's no longer in `.gitignore`)~~

In the deployment workflow, I use a different action that makes deployments of whole directorries easier. I have [tested it](https://github.com/Jartreg/tscript/runs/3766628510?check_suite_focus=true) but I don't have much experience with this action. I'm confident it will work well but that's something to look out for.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -38,17 +38,16 @@ jobs:
         run: npm test
 
       # Deploy to GitHub Pages
-      - name: Commit distribution artifacts
+      - name: Prepare deployment
         run: |
-          git checkout --orphan dist
-          git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com"
-          git config --local user.name "github-actions[bot]"
-          git reset
-          git add -f ./distribution/index.html
-          git commit -m "Deploy site"
-      - name: Push changes
-        uses: ad-m/github-push-action@master
+          rm ./out/tscript.js
+          mkdir ./out/distribution
+          mv ./src/ide/redirect.html ./out/distribution/index.html
+      - name: Deploy to GitHub Pages
+        uses: JamesIves/github-pages-deploy-action@4.1.5
         with:
-          github_token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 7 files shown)


.gitignore
```diff
@@ -1,2 +1,2 @@
 node_modules/
-distribution/
```
manuel-fischer, line 2: Maybe keep this in, because otherwise one might accidentally upload the old distribution folder from the local working directory.

.github/workflows/cd.yml
```diff
@@ -38,17 +38,16 @@ jobs:
...
         run: |
-          git checkout --orphan dist
-          git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com"
-          git config --local user.name "github-actions[bot]"
-          git reset
-          git add -f ./distribution/index.html
-          git commit -m "Deploy site"
-      - name: Push changes
-        uses: ad-m/github-push-action@master
+          rm ./out/tscript.js
```
manuel-fischer, line 43: Why does it get removed?

manuel-fischer (approved):

.gitignore
Jartreg: Yeah, I totally see that happening 👍

.github/workflows/cd.yml
Jartreg: This is the build output from `rpm run interop`. It's not used in the IDE and is just produced to test that it builds without errors. It's not uploaded currently and I don't really see a good reason to change that.

PR #98, opened 2021-10-05
"Comment" in EBNF causes problem

manuel-fischer:
- Maybe the EBNF Syntax could be extended by comments introduced with "#"
- Also the syntax of various code in the documentation could be checked inside tests

### DIFF
--- a/src/doc/language.ts
+++ b/src/doc/language.ts
@@ -2157,7 +2157,7 @@ print(r > s);                           # prints true, as the first comparison t
 			<p>
 			A function call is an expression with the following syntax:
 			<ebnf>
-				function-call = expression "(" argument { "," argument } ")" ";" ;          # example("example");
+				function-call = expression "(" argument { "," argument } ")" ";" ;          '# example("example");'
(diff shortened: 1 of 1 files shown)


TGlas:
By the same token, I think that the EBNF syntax is not consistent. In the lecture, a sequence requires commas as separators (and that's consistent with Wikipedia). For brevity and readabilty the commas are dropped in many places. An automatic syntax check would force us to fix that.

Most code examples in the documentation are checked for (non-semantic) errors by running the code in the background. Having syntax checks for EBNF would be a useful addition to that. I don't know whether that's possible, but maybe these checks should not run in the browser, but as a github action.

manuel-fischer:
It might be useful to create a new issue concerning syntax checking of code in the documentation.

The specific case of the error in the "function-call" documentation page might be important to be fixed before the lectures start, because the page itself cannot be viewed due to the error.

TGlas:
Okay, point taken. This is a small thing that should go first. You are right, I'll move my comment to a new issue.

PR #99, opened 2021-10-05
Fix tokenize throwing when no tokens are found

Jartreg:
The `tokenize` function inside `search.ts` may throw a `TypeError`. The error occurs because `words` is `null` if `String#match` doesn't find any tokens. This is fixed by returning an empty array if no tokens are found.

a screenshot of the TypeError in the console


TGlas:
Good catch!

PR #101, opened 2021-10-05
Use query parameters instead of the hash value for URLs

Jartreg:
Similar to #97, this is a change to URLs. The IDE and especially the documentation currently use the hash value and the entire search string for navigation. This puts some limitations on us because we can't use hashes for _actual_ hash navigation (e.g. skipping to certain sections on a doc page) and it gets hard when there are multiple parameters at play (I'm looking at you, `#search/what/is/this`!).

What a coincidence it is that there's something better suited for this: ✨ Query Parameters ✨!

This PR intends to change the URLs used by the IDE and documentation to work like this:

Old URL | New URL
--------|--------
`index.html?doc` | `index.html?doc=`
`index.html?doc#/<path>` | `index.html?doc=<path>`
`index.html?doc#search/<term1>/<term2>` | `index.html?doc=search&q=<terms>`
`index.html?run#<url>` | `index.html?load=<url>`

### Regarding the implementation

Old URLs of course still work and are redirected (see `translateLegacyURL` in `src/ide/index.ts`).

I didn't change the URLs used as `href`s by errors, because there are a lot of places where they are used, making it easy to introduce bugs. Instead, when clicking on an error message, `showdoc` automatically translates the URL.

`navigation.ts` contains new utilities for navigation, including the almighty `navigate`-function. Clicks on links are automatically intercepted, other places should use the `navigate`-function. I introduced `IPageController`s, whose job it is to show the the page they are responsible for. There are two new files for the `IPageController`s of the IDE and the documentation, which are currently just wrappers for the existing code. They should, however, grow and take over more of the work if `ide.ts` and `doc.ts` are split up or refactored.

Navigation now works like this:
- When loading the page, a closure in `index.ts` determines which `IPageController` to use (either the IDE or the documentation)
- `navigation.ts` creates a new instance of it which then builds the contents of the page.
- When navigating within an `IPageController` (e.g. navigating from one doc page to another), the `IPageController` tries to handle that (i.e. no browser navigation, as before!). The history and address bar are updated using [`history.pushState()`](https://developer.mozilla.org/en-US/docs/Web/API/History/pushState).

I didn't implement navigating from one `IPageController` to another, because this currently never happens. Instead, it just defaults to a full browser navigation.


src/ide/navigation.ts
```diff
@@ -0,0 +1,116 @@
...
+		window.addEventListener("beforeunload", (event) => {
+			if (currentController?.checkUnsavedChanges?.()) {
+				event.preventDefault();
+				event.returnValue = "";
+			}
+		});
+	}
+}
+
+function tryHandleNavigation(newUrl: URL): boolean {
```
TGlas, line 55: Could you document the idea behind this function? It becomes clear only when checking the navigate function.

TGlas (commented):
Maybe it is overkill to introduce an inheritance chain for a single switch, but I can see that this is much better for separating IDE and documentation. Do you consider this PR ready for merging, or do you plan to edit more?

TGlas:
I don't have a strong opinion on moving to `index.ts` or not. If you find it cleaner then go ahead :)

PR #109, opened 2021-10-28
alert+confirm+prompt dialogs replaced with own modal dialogs

### LINKED ISSUE #106: print message not showing before alert
### Description
The TScript code

    print(1);
    alert(2);
    print(3);

works fine in Firefox and Chromium (at least on Ubuntu), but it behaves strangely in Chrome and Opera on Windows. While the alert message is active, the "1" from the first print statement should be visible in the message area. After closing the message box, the "3" should follow. In the above browsers, both print messages are rendered only after closing the message box.

### To Reproduce
The above can easily be verified with the right OS/browser combination, or using a platform like browserling. Start the TScript IDE from [here](https://tglas.github.io/tscript/index.html). Type the above code and run the TScript program. When the alert shows up, no print results are visible. After closing the message box, both print results show up.

### Possible Remedies
Until now, none of my attempts at a "simple" fix worked. Others seem to face [similar problems](https://stackoverflow.com/questions/8840580/force-dom-redraw-refresh-on-chrome-mac). The available fixes are dirtry tricks, which may stop working any time. There does not seem to be a JS API to force a page update (reflow or redraw). On the other hand, the DOM modification of adding the print message should do so anyway, but it does not take effect until the browser decides to dedicate its cycles to the task. The problem is that the modal dialog seems to keep it from doing so, for no reason (the JS call should block, not the entire browser engine).

A possible fix may be to show a tgui-dialog (like the configuration dialog) instead of the JS built-in alert/confirm/prompt dialogs. That would keep the page alive and leave more than enough time to render the update. However, this would require a reorganization of the corresponding interpreter services, which are currently assumed to be blocking functions. The interpreter would need to wait idle for an event triggered by a callback, similar to what it does in "waiting" mode. The callback would need to set the return value (in case of confirm and prompt) and trigger the continuation of the interpreter loop.

TGlas:
This is my attempt to fix #106. The JS-native `alert`, `confirm` and `prompt` dialogs are replaced with tgui-based dialogs. According to my tests with various browser and OS combinations, this results in correct interpretation of the test program

    print(1);
    alert(2);
    print(3);

and similarly with `confirm` and `prompt`. The price is slightly increased complexity of the interpreter, which now has a "dialog" mode. That mode is similar to "waiting", but instead of continuing at a fixed time, it waits for the result from the modal dialog. Also, the alert/confirm/prompt services are now asynchronous: instead of returning a value they take a callback function as an additional argument.

I tested the following platforms:

- Firefox on Linux
- Chromium on Linux
- Chrome on Windows
- Opera on Windows

I'd appreciate further browser tests, and also a code review.

### DIFF
--- a/src/ide/ide.ts
+++ b/src/ide/ide.ts
@@ -822,92 +822,98 @@ export let ide = (function () {
 				module.addMessage("print", msg);
 				module.interpreter.flush();
 			};
-			module.interpreter.service.alert = function (msg, reportValue) {
-				let dlg = tgui.msgBox({
-					title: "",
-					prompt: msg,
-					icon: tgui.msgBoxExclamation,
-					buttons: [{ text: "Close", isDefault: true }],
-					enterConfirms: true,
-					onClose: () => {
-						reportValue();
-						return false;
-					},
+			module.interpreter.service.alert = function (msg) {
+				return new Promise((resolve, reject) => {
+					let dlg = tgui.msgBox({
+						title: "",
+						prompt: msg,
+						icon: tgui.msgBoxExclamation,
+						buttons: [{ text: "Close", isDefault: true }],
+						enterConfirms: true,
+						onClose: () => {
+							resolve(null);
+							return false;
+						},
+					});
 				});
 			};
-			module.interpreter.service.confirm = function (msg, reportValue) {
-				let value = false;
-				let dlg = tgui.msgBox({
-					title: "Question",
-					prompt: msg,
-					icon: tgui.msgBoxQuestion,
-					buttons: [
-						{
-							text: "Yes",
-							isDefault: true,
-							onClick: () => {
-								value = true;
-								return false;
+			module.interpreter.service.confirm = function (msg) {
+				return new Promise((resolve, reject) => {
+					let value = false;
+					let dlg = tgui.msgBox({
+						title: "Question",
+						prompt: msg,
+						icon: tgui.msgBoxQuestion,
+						buttons: [
+							{
+								text: "Yes",
+								isDefault: true,
+								onClick: () => {
+									value = true;
+									return false;
+								},
 							},
-						},
-						{
-							text: "No",
-							isDefault: false,
-							onClick: () => {
-								value = false;
-								return false;
+							{
+								text: "No",
+								isDefault: false,
+								onClick: () => {
+									value = false;
+									return false;
+								},
 							},
+						],
+						enterConfirms: true,
+						onClose: () => {
+							resolve(value);
+							return false;
 						},
-					],
-					enterConfirms: true,
-					onClose: () => {
-						reportValue(value);
-						return false;
-					},
+					});
 				});
 			};
-			module.interpreter.service.prompt = function (msg, reportValue) {
-				let input = tgui.createElement({
-					type: "input",
-					classname: "ide-prompt-input",
-					properties: { type: "text" },
-				});
-				let value = null;
-				let dlg = tgui.createModal({
-					title: "Input",
-					scalesize: [0.2, 0.15],
-					minsize: [300, 150],
-					buttons: [
-						{
-							text: "Okay",
-							isDefault: true,
-							onClick: () => {
-								value = input.value;
-								return false;
+			module.interpreter.service.prompt = function (msg) {
+				return new Promise((resolve, reject) => {
+					let input = tgui.createElement({
+						type: "input",
+						classname: "ide-prompt-input",
+						properties: { type: "text" },
+					});
+					let value = null;
+					let dlg = tgui.createModal({
+						title: "Input",
+						scalesize: [0.2, 0.15],
+						minsize: [300, 150],
+						buttons: [
+							{
+								text: "Okay",
+								isDefault: true,
+								onClick: () => {
+									value = input.value;
+									return false;
+								},
 							},
-						},
-						{
-							text: "Cancel",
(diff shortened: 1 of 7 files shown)


Jartreg:
I have yet to test it, but I had an idea after reading through the changes:
How about returning `Promise`s from the service methods rather than passing a callback?

Basically, in a service method:
```ts
return new Promise((resolve) => {
	// show the dialog
	// when done:
	resolve(/* whatever value */);
});
```

And in the lib:
```ts
this.service.alert(s).then((value) => /* ... */);
```

I think this could have some benefits in the future:
* Using async/await in service implementations
* Enabling the interpreter to natively wait for promises at some point (could be used in other places as well),
  **meaning it wouldn't have to poll**

TGlas:
Not having to poll is definitely a good thing. I was stumbling across promises myself in the process. As far as I can see, they cannot turn asynchronous code into synchronous code. Since that was by main concern, I did not follow up on promises, but I did not think about the option to avoid polling. In the end, the interpreter thread is polling anyway.

Maybe polling can also be avoided for waiting (TScript function `wait`). Actually, there is a difference between waiting for a point in time in the future and waiting for the completion of a modal dialog: the first one can be interrupted by stopping the program, the second one cannot. Maybe that should be enabled, by the dialog only blocking the panels, not the toolbar and the corresponding hotkeys (F10 by default).

I don't know if the possibility to interrupt would make dealing with a promise more difficult. It seems that JS does not offer the option to cancel a promise. Of course, that could be emulated. Not sure... maybe in that scenario polling is the easier solution?

Jartreg:
There's definitely a lot to consider when thinking about how promises could be used throughout the interpreter, but I'm not talking about switching everything to promises right now in this PR 😄.

I'd rather just want to suggest using them _as an alternative to callbacks_, because this _could_ lead to benefits later. If you'd just replace the current use of callbacks with promises, nothing would change about polling anyway 🤷‍♂️.

(As an additional note regarding the interruption of promises: when people were discussing how to cancel a `fetch` request, they added [`AbortController`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController), which can also be used outside of `fetch`.)

src/ide/ide.ts
```diff
@@ -820,13 +823,97 @@ export let ide = (function () {
 				module.interpreter.flush();
 			};
 			module.interpreter.service.alert = function (msg) {
-				alert(msg);
+				return new Promise((resolve, reject) => {
+					let dlg = tgui.msgBox({
+						title: "",
+						prompt: msg,
+						icon: tgui.msgBoxExclamation,
+						buttons: [{ text: "Close", isDefault: true }],
```
Jartreg, line 831:
```suggestion
						buttons: [{ text: "Okay", isDefault: true }],
```

I think I'd rather prefer it without an icon and with a different button label, since that's how most browsers show alerts 🤷‍♂️

src/ide/ide.ts
```diff
@@ -622,6 +622,8 @@ export let ide = (function () {
 				else module.programstate.stepping();
 			} else if (module.interpreter.status === "waiting")
 				module.programstate.waiting();
+			else if (module.interpreter.status === "dialog")
+				module.programstate.dialog();
```
Jartreg, line 626:
```suggestion
				module.programstate.waiting();
```

Why not just show "program is waiting" here? I don't think the dialog state is too special that we should expose it to the user (it's almost hidden behind the backdrop anyway). In addition, I'd find it more consistent if we indicated any kind of waiting as "program is waiting".

src/lang/tscript-lib/core.ts
```diff
@@ -957,43 +957,60 @@ export const core = {
 		},
 		alert: function (text) {
 			let s = TScript.toString.call(this, text);
-			if (!this.service.documentation_mode && this.service.alert)
-				this.service.alert(s);
+			if (!this.service.documentation_mode && this.service.alert) {
+				this.status = "dialog";
```
Jartreg, line 961:
```suggestion
				this.status = "dialog";
				this.service.statechanged?.(false);
```

You probably need to call `statechanged`, as it's currently not showing any state change when opening a dialog.

src/lang/tscript-lib/core.ts
```diff
@@ -957,43 +957,60 @@ export const core = {
...
 				value: { b: null },
 			};
 		},
 		confirm: function (text) {
 			let s = TScript.toString.call(this, text);
-			let ret = false;
-			if (!this.service.documentation_mode && this.service.confirm)
-				ret = this.service.confirm(s);
+			if (!this.service.documentation_mode && this.service.confirm) {
+				this.status = "dialog";
```
Jartreg, line 978:
```suggestion
				this.status = "dialog";
				this.service.statechanged?.(false);
```
Same as in `alert`

src/lang/tscript-lib/core.ts
```diff
@@ -957,43 +957,60 @@ export const core = {
...
-					return {
-						type: this.program.types[Typeid.typeid_string],
-						value: { b: ret },
-					};
-			} else
-				return {
-					type: this.program.types[Typeid.typeid_string],
-					value: { b: "" },
-				};
+				this.status = "dialog";
```
Jartreg, line 995:
```suggestion
				this.status = "dialog";
				this.service.statechanged?.(false);
```
Same as in `alert` and `confirm`

Jartreg (changes requested):

PR #114, opened 2021-11-24
fix path sanitization

Freax13:
#110 reintroduced the XSS vulnerability fixed by #108. `showPage` must also sanitize the path that is passed on internally, not only the user visible URL.

@Jartreg Can you verify that this pr doesn't break anything?

### DIFF
--- a/src/escape.ts
+++ b/src/escape.ts
@@ -1,0 +1,13 @@
+export function escapeHtmlChars(s) {
+	let ret = "";
+	for (let i = 0; i < s.length; i++) {
+		let c = s[i];
+		if (c == "\n") ret += "<br/>";
+		else if (c == '"') ret += "&quot;";
+		else if (c == "<") ret += "&lt;";
+		else if (c == ">") ret += "&gt;";
+		else if (c == "&") ret += "&amp;";
+		else ret += c;
+	}
+	return ret;
+}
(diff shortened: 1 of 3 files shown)


Jartreg (changes requested):
How about we approach this differently... I don't think it's a bug that should be fixed here in `index.ts` since its duty should basically just be to pass on url parameters.

I'd rather suggest that we sanitize the tokens when they are displayed:
https://github.com/TGlas/tscript/blob/f34a088587a51e4a4cd84fdf1f6153aa4adecbb3/src/ide/doc.ts#L672-L675

Escaping characters like `<`, `>` and `&` in `key[i]` before adding it to `html` should resolve this issue at the root. _Ideally_ we'd use `Text` nodes to avoid having to deal with sanitization altogether, but that would require too much refactoring.

What do you think?

Freax13:
Fair enough, we could reuse the escape function already in the code base: https://github.com/TGlas/tscript/blob/f34a088587a51e4a4cd84fdf1f6153aa4adecbb3/src/eval/index.ts#L138-L150

I'd like to note that JavaScript already has a (deprecated) function called `escape` used for escaping url queries and that function is also used within the code base: https://github.com/TGlas/tscript/blob/f34a088587a51e4a4cd84fdf1f6153aa4adecbb3/src/ide/ide.ts#L1151

So if we pull the function out of `eval/index.ts` and reuse it in `ide/doc.ts`, I'd like to suggest renaming the function eg. to `escapeHtmlChars` to avoid confusion between JavaScript's `escape` and our `escape`.

TGlas:
Good point. I imported the code in `eval/index.ts` from a different source without putting too much effort into making it consistent with the rest. Feel free to rename and move the function around.

TGlas:
Thanks, looks good!

PR #118, opened 2022-01-07
added tutorial and unit tests for all exercises

duckmopsi:


src/ide/ide.ts
```diff
@@ -2524,7 +2524,7 @@ export let ide = (function () {
 			e = createTypedEvent("canvas.KeyboardEvent", e);
 			module.interpreter.enqueueEvent("canvas.keyup", e);
 		});
-		/*
+
```
Jartreg, line 2527: Was it intentional to uncomment this?

duckmopsi: Not sure tbh, I intended to only uncomment the code that's necessary for the tutorial.

Jartreg: Well, I'm asking because the title only says something about unit tests and this doesn't seem to be related...?

duckmopsi: Yes, I thought the tutorial had already been added but there hasn't been a PR to do that. Changed the title accordingly.

TGlas:
Many many thanks!

Issue #123, opened 2022-10-12
IDE: Inconsistent theming on first load


pfz4:
### Description
Hello, I've noticed that the IDE Theming is inconsistent on the first load. The light theme is used when the browser is preferring a dark theme.

### To Reproduce
1. Open an empty session in the browser of your choice, that is preferring dark themes, and go to the IDE. => The IDE is using the "light" theme.
2. Open the IDE-Settings and change the theme to "light" or "dark" => The IDE applies the selected Theme
3. Change the theme back to "System Default" => The IDE applies the browser-preferred ("dark") theme.

The IDE should use the browser-preferred theme from the start.

### Why Does This Error Occur?
The "light" theme is defined as the default theme in [`tgui.ts`:33](https://github.com/TGlas/tscript/blob/44f04d7fbf87b3a2c7ceb0745d65719238b35d89/src/ide/ide.ts#L33).
Files using `tgui.ts` (like `ide.ts` and `doc.ts`) define the theme as "default" (use the browser-preferred theme). These Modules only update the theme on load (using `tgui.setTheme(...)
;`) when the config is available from local storage. 

### Fixing the Issue
The first option would be to set the theme in the modules using `tgui.ts` without checking if the config is valid (falling back to the default state of the `theme` variable) (See [`ide.ts`](https://github.com/TGlas/tscript/blob/44f04d7fbf87b3a2c7ceb0745d65719238b35d89/src/ide/ide.ts#L1422)). I don't like that approach because it requires every module that uses `tgui.ts` being touched and does not prevent that error from happening in future modules.

The second option I can think of would be to call `module.setTheme("default")` before returning `module` in [`tgui.ts`:1716](https://github.com/TGlas/tscript/blob/2c4dcb3ec8a714927e3e5519158b5dd3882a7329/src/ide/tgui.ts#L1716), setting the theme to the browser-preferred theme.
Sadly, it does not work to simply change the default value of the `theme` variable in `tgui.ts` because some logic is required to fetch the browser-preferred theme.

Thanks for taking the time to look at this issue. I would be happy to create a PR implementing the preferred option :)

TGlas:
Thanks for the report. I do not have a strong opinion on this, but I do follow your logic and therefore think that the second option is better. Please create a PR.

PR #124, opened 2022-10-14
IDE: Use operating system theme on first load

pfz4:
Issue: [https://github.com/TGlas/tscript/issues/123](https://github.com/TGlas/tscript/issues/123)

As proposed in the issue, the IDE now uses the operating system theme when no theme is saved in the local Storage.

I've decided to **not** change the constructor of `tgui.ts`.
I've overlooked that the theme needs to be set after the DOM is constructed. To implement my suggested "Option 2", I would have ve to register another event listener to "DOMContentLoaded". I decided not to do that to avoid risking interfering with other registered event handlers.

I've now changed the `doc.ts` and `ide.ts` files to always set the theme to "default" if no theme is specified in the local Storage.

### DIFF
--- a/src/ide/doc.ts
+++ b/src/ide/doc.ts
@@ -22,15 +22,15 @@ export const doc = (function () {
 	let docpath = "";
 	let doctree: any = null;
 
-	// check if dark theme is activated
+	// load theme from localStorage or use the theme of the operating system
 	function loadTheme() {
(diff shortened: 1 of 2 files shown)


pfz4:
Oh, I'm sorry. Should be pretty now

TGlas:
Thanks!!

PR #127, opened 2023-11-16
Added slice function

TheRealFloatDev:
I noticed that the Array type was lacking a slice function, so I implemented it.

It has been added in the core and core.tscript.ts as well as the documentation.

It has also been tested and is working fine.
It also includes necessary checks for the parameter types.

### DIFF
--- a/src/doc/language.ts
+++ b/src/doc/language.ts
@@ -4030,6 +4030,9 @@ print(infiniteR.isFinite());        # prints false
 				The <code class="code">function size()</code> returns the size of
 				the array, which is the number of items.
 			</td></tr>
+			<tr><th>slice</th><td>
+				The <code class="code">function slice(start, end)</code> returns the content of the given range in the array.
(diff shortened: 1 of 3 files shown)


TGlas:
Why do we need a slice function if the subscript operator does the job?

var a = [4,5,6,7,8,9];
print(a[2:4]);    # prints [6,7]

Maybe I should make that more clear by adding an example to the Array documentation.

TheRealFloatDev:
I would add it because students who are already experienced programmers are used to having the slice function, since it exists in nearly all modern languages and also I believe it is easier to understand than the subscript operator.

TGlas:
I see that point, and it does no harm to have two equivalent mechanisms.

TGlas:
Same comment as in the other branch: please `npm run prettier` to make the checks pass.

TGlas:
Thanks!

PR #128, opened 2023-11-16
Additional String functions

TheRealFloatDev:
I also noticed that the String type is lacking some basic functionality, more precise I added:

- toLowerCase()
- toUpperCase()
- replace()

All functions are also added in the docs


TGlas:
Thanks, that's really useful. However, please disallow regular expressions in "replace" to make the method more beginner friendly.

TheRealFloatDev:
I have now removed regex support from the function.

TGlas:
Thanks! One more small thing: Please run `npm run prettier` before to make the checks succeed.

TGlas:
Thanks a lot.

PR #129, opened 2023-11-18
Added array indexOf function

TheRealFloatDev:
I have now also added the indexOf function to the Array Type. I think it is one of the basic functions an array type needs and would have also been useful for the Recursion exercise.

I have not added the option to also pass a starting index, because I think this is too complicated for beginners, but I can add it if you want.


TGlas:
Actually, this one is missing on purpose. I agree that it is basic and very useful, but I want people to write it by themselves. This is actually somewhat related to runtime considerations. I want students to "experience" the loop, i.e., linear complexity of the search, which is in contrast to constant complexity of a key lookup in a dictionary (hash table). These topics will play an important role in the second half of the CS 1 lecture.

PR #130, opened 2024-02-02
Studienprojekt TScript Editor Draft

esatbayhan:
Changes:
- Refactoring mainly ide.ts and tgui.ts
    - Added comments
    - Split up to files
    - Reduce exports
- Updated CodeMirror 5 to 6
    - Wrap all CodeMirror code inside TScriptEditor
    - Migrate CodeMirror code matching Version 6
    - Maintain all functionalities (syntax highlighting, debugging, etc.)
- Added multi document editor
    - Multiple documents can be opened inside panels
    - Error messages are matched to corresponding files and opens them


TGlas:
The code looks very good! The refactoring also helps a lot, both in terms of manageable file size and type safety.
I tried the code and it works mostly well. There are still a few glitches. Could you address the following points?

 * The documentation for the "Source Code Editor" should say something about the multi-document editor.
 * Ctrl+D does not work (block comment/uncomment code); seems to be used differently by CodeMirror6.
 * The table of contents of the documentation (tree view on the left) should use sans font, not monospace.
 * It would be nice if all editor panels would precede the other panel types when arranged.
 * Sometimes outdated versions of an included file seem to be used. However, I cannot reproduce the problem... so ignore for now. Also not sure whether that's already a problem in the original code base.
 * The directory structure `src/ide/ide/...` is somewhat odd. Could you avoid the nested directory and move the new files one level up? Probably needs renaming `src/ide/ide/index.ts`.

PR #138, opened 2025-03-18
Cleanup/Typings of Editor Options and Themes

Jartreg:
This PR primarily contains two changes:

- IDE/Editor themes have a type
- The options passed to the Editor constructor now have a type

These changes are related because the editor needs a theme type for its options and makes use of this type in multiple places.

## Cleanup of Theme Handling

While adding theme types, I cleaned up how IDE themes are handled and theme changes are propagated through the IDE.
Previously, there were **two different** variables keeping track of the current theme:

https://github.com/TGlas/tscript/blob/3d3ccb4a8bdeb4d7a4b3f8699768ce374ffb4751/src/ide/elements/dialogs.ts#L8
https://github.com/TGlas/tscript/blob/3d3ccb4a8bdeb4d7a4b3f8699768ce374ffb4751/src/ide/tgui/index.ts#L16

I encapsulated all state related to themes in `src/ide/tgui/theme.ts`. One remaining annoyance: the configured theme is still loaded in two different places, for the IDE and the docs respectively.

In addition, theme changes were propagated in a spaghetti-ish way. Editors were notified **twice**:

1. The configuration dialog iterated over editors and updated their themes.
1. `tgui/index.ts` iterated over all panels and called `onThemeChange` &rarr; the editor panel iterated over editors and updated their themes.

I replaced that with a simple subscription mechanism. `EditorCollection` subscribes to theme changes and notifies its editors. The editor used for reference solutions previously wasn't updated and I didn't change that.

## Typings for Editor Options

There's now a type for the options of the Editor constructor. This needed types for language identifiers, language definitions and theme definitions. Adding these types actually caused TypeScript to find two bugs in `language.ts`!
There are some places where I'm not sure I fully understood what the code is doing. I commented on these places below to clear things up.

### DIFF
--- a/src/ide/editor/language.ts
+++ b/src/ide/editor/language.ts
@@ -9,7 +9,12 @@ import { Document } from "./document";
 // ideally before creating editor instances.
 //
 
-let language_definitions = {
+interface LanguageDefinition {
+	comments: { line?: string; begin?: string; end?: string };
+	highlight: Record<string, string>;
+}
+
+const language_definitions = {
 	plain: {
 		comments: {},
 		highlight: {
@@ -92,7 +97,9 @@ let language_definitions = {
 			comment: "gray comment; \n plain start;",
 		},
 	},
-};
+} satisfies Record<string, LanguageDefinition>;
+
+export type LanguageIdentifier = keyof typeof language_definitions;
 
 // Language definition, consisting of comments and a syntax scanner.
 // The special property of the scanner is that it can enter in the middle of a token.
@@ -110,11 +117,11 @@ export class Language {
 	table: Array<Uint32Array> = []; // transition table
 	num_pending: Array<number> = []; // state -> # pending characters
 
-	public constructor(name: string) {
-		if (!language_definitions.hasOwnProperty(name))
+	public constructor(name: LanguageIdentifier) {
+		if (!Object.hasOwn(language_definitions, name))
 			throw "unknown highlighting language";
-		let definition = language_definitions[name];
-		let rules = definition.highlight;
+		const definition: LanguageDefinition = language_definitions[name];
+		const rules = definition.highlight;
 
 		this.linecomment = definition.comments.line
 			? Document.str2arr(definition.comments.line)
@@ -128,7 +135,7 @@ export class Language {
 
 		// count and enumerate named rules
 		let num = 1;
-		let name2state = {};
+		const name2state: Partial<Record<string, number>> = {};
 		name2state["start"] = 0;
 		this.table.push(new Uint32Array(99));
 		this.num_pending.push(0);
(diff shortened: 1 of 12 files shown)


src/ide/editor/theme.ts
```diff
@@ -5,7 +5,48 @@
...
+			string,
+			string,
+			string,
+			string,
+			string,
+			string,
+			string,
+			string,
+			string
+		];
```
Jartreg, line 35: I unfortunately didn't find any reason why this array has this length, so I kept it as a tuple of 10 strings.

src/ide/editor/language.ts
```diff
@@ -191,7 +198,7 @@ export class Language {
 				let set = new Set<number>();
 				while (true) {
 					let begin = rule.charCodeAt(pos);
-					if (begin === " ") break;
+					if (begin === 0x20) break; // Space
```
Jartreg, line 201: `charCodeAt` returns a number. This might be a behavioural change since the loop was never exited here. Debugging indicated that spaces might've previously been caught by `if (rule[pos] !== "+") break;` below.

src/ide/editor/language.ts
```diff
@@ -264,7 +271,7 @@ export class Language {
 					let keywords = rule
 						.substring(pos, rule.length - 1)
 						.split(" ");
-					let prefixes = new Map<Uint32Array, number>(); // map prefix to state index
+					const prefixes = new Map<string, number>(); // map prefix to state index
```
Jartreg, line 274: TypeScript complained in line 283 (`prefixes.has(prefix)`), that `prefix` is of type `string` while this map expects keys of type `Uint32Array`. I'm pretty sure that the key type should simply be `string`…?

src/ide/editor/theme.ts
TGlas: That's a rather quick and dirty design. The ten strings hold CSS color values like "#000". They should probably be named rather than accessed by index. Index 0 is "normal" style (no highlighting), index 1 is a thee keyword highlight color, ..., index 9 is a comment. If you feel like refactoring the collection then please go ahead.

src/ide/editor/language.ts
TGlas: I agree. At that point the (partial) keywords are represented as strings, as is evident from `kw` and `prefix`.

TGlas:
Thanks for looking into this. The current state is the result of two independent restructuring processes (introducing 1. multiple tabs and 2. a new editor), none of which took the other into account. As a result, the glue code in between is all but clean, and I am not surprised that you found some Spaghetti in there.

TGlas:
Please let me know when you consider the PR ready for merging.

Jartreg:
I added a new enum for highlight color values. This doesn't change anything about how the code works (the values remain the same), but it's maybe a little more clear what the values mean.
I'd consider this ready for merging.

PR #140, opened 2025-04-05
Decoupling the Standalone Page from the IDE

Jartreg:
Currently, standalone exports initialize the entire IDE and open Editors for each included file, just so the interpreter can load the code.

This PR does the following:

- The last commit decouples the standalone page from the IDE by separating common logic for the creation of the interpreter and canvases
  - Canvases for the turtle/canvas namespaces are now newly created for each run and removed with the interpreter. They handle resizes by observing their container element.
  - Objects belonging to the interpreter (the canvases) are grouped in a new class `InterpreterSession`
- Loading of included files is moved from the parser into a callback
- The remaining commits add (a lot of) types and refactors to enable this

## Accidental Bugfixes/Features

The parser previously loaded included files from localStorage. In environments where localStorage is not available, using the `include` statement would've probably lead to a crash (NodeJS, exam evaluation?).

To export a program using includes, it was previously necessary to open all included files in editors (and I couldn't find any documentation on this). Now, the export command is able to automatically determine which files are included and only adds those to the output.

## Regarding the Merge

Since this PR touches a lot of files, I split some changes into multiple commits and gave (hopefully) good descriptions on what is changed. Is there a specific reason why you like to squash PRs? If you'd like, I could still combine some of these commits into one.

### DIFF
--- a/src/lang/interpreter/program-elements.ts
+++ b/src/lang/interpreter/program-elements.ts
@@ -1,0 +1,62 @@
+import { ParserPosition } from "../parser";
+import { Interpreter } from "./interpreter";
+
+interface ProgramElementBase<Type extends string> {
+	readonly petype: Type;
+
+	/** The position where this program element starts in the source code */
+	where?: ParserPosition;
+
+	/**
+	 * Execute this program element
+	 *
+	 * @returns whether the program element is done. If false, it must be called again with an incremented instruction pointer.
+	 */
+	step(this: Interpreter): boolean;
+
+	/**
+	 * Determines whether the program element would be done after the next call to {@link step}.
+	 *
+	 * This controls when the debugger stops when stepping through a program.
+	 * Since the debugger wants to stop just before the PE's last instruction,
+	 * it halts when this method returns true.
+	 *
+	 * Some program elements don't want the debugger to halt on them and always return false
+	 * (e.g. by using `simfalse`).
+	 */
+	sim(this: Interpreter): boolean;
+}
+
+export type StepFn = ProgramElementBase<any>["step"];
+export type SimFn = ProgramElementBase<any>["sim"];
+
+/** like a main function, but with more stuff */
+export interface ProgramRoot extends ProgramElementBase<"global scope"> {
+	/** children in the abstract syntax tree */
+	children: any[];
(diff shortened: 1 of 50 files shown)


src/lang/parser/lexer.ts
```diff
@@ -216,23 +230,22 @@ export class Lexer {
...
-					else state.skip();
-					tok = { type: "operator", value: op, code: op, line: line };
-				}
 			}
-			if (tok === null) {
-				let type: any = null;
+
+			if (op && op !== "!") {
+				tok = { type: "operator", value: op, code: op, line: line };
+			} else {
```
Jartreg, line 236: This restructuring was necessary so that TypeScript knows that tok is definitely assigned after the else branch ("all the rest").

src/lang/interpreter/program-elements.ts
```diff
@@ -0,0 +1,62 @@
...
+	 * Determines whether the program element would be done after the next call to {@link step}.
+	 *
+	 * This controls when the debugger stops when stepping through a program.
+	 * Since the debugger wants to stop just before the PE's last instruction,
+	 * it halts when this method returns true.
+	 *
+	 * Some program elements don't want the debugger to halt on them and always return false
+	 * (e.g. by using `simfalse`).
+	 */
+	sim(this: Interpreter): boolean;
```
Jartreg, line 27: I hope I got this right. I was initially confused that some PEs always return `false` here. So I think despite its name this method is actually only used to control when the debugger stops…

TGlas:
First of all: thanks a lot for you efforts! The export feature was previously very quick and dirty (creating lots of unneeded elements and hiding them again), so it needed cleanup real badly.

TGlas:
Regarding your question on squash PRs: I don't have a strong opinion on how to merge. "Squash and merge" is simply the default option offered by github. Of course, I can also do a "normal" merge commit.

src/lang/interpreter/program-elements.ts
TGlas: Yes, you got that right. The expectation for `sim` is that it returns the same value as the next call to `step` will do, but without executing the step. As far as I remember, the only purpose is a nicer behavior of the debugger.

PR #141, opened 2025-05-03
TreeControl refactor: ES6 Class and async info callback

ole0811sch:
This pull request refactors the `TreeControl` object returned by `createTreeControl` into a fully-typed class and allows the `TreeDescription.info` callback (which describes the structure of the tree) to be asynchronous. It also improves code documentation for the expected behavior of `TreeDescription.info`.


TGlas:
Thanks a lot, that looks good!

PR #142, opened 2025-05-15
Reduce asynchronous UI code caused by TreeControl

Jartreg:
@ole0811sch: If I understand correctly, you need the info callback to be able to return promises.

I think that the UI code should have as few async paths as possible, since it increases the potential for bugs.
For example, I'm not sure whether it would be fine if the async version of `updateControls` is called while a previous call hasn't completed yet.

To mitigate this, I reverted the callers of `TreeControl#update` to be synchronous. Instead, `TreeControl` now manages asynchronous updates itself.

on `update(…)`:

- if no asynchronous update is running, a new one starts
- if there is one, `TreeControl` remembers to start a new one once the current has completed

**Note:** As before, there is no handling for errors that occur during the update

@TGlas: While I previously [advised against using private class members](https://github.com/TGlas/tscript/pull/96#discussion_r718310758), I've realized that I was totally wrong on the reasons. TypeScript compiles them away anyway, and ~3 years later, [browser support](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties#browser_compatibility) has changed as well. So it would still work if the compilation target was increased.

Are you fine with having both `private property` and `#property` in the project? Or should I stick to TypeScript's modifiers?

### DIFF
--- a/src/ide/doc.ts
+++ b/src/ide/doc.ts
@@ -664,7 +664,7 @@ export const doc = (function () {
 		return [node, parent, parentpath, index];
 	}
 
-	module.setpath = async function (path) {
+	module.setpath = function (path) {
 		if (!path) path = "";
 		if (path.length > 0 && path[0] == "#") path = path.substr(1);
 
@@ -712,7 +712,7 @@ export const doc = (function () {
 			module.dom_content.innerHTML = html;
 			module.dom_content.scrollTop = 0;
 			docpath = "";
-			await doctree!.update(docinfo);
+			doctree!.update(docinfo);
 		} else {
 			try {
 				let data = getnode(path);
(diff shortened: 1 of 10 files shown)


ole0811sch:
I hadn't considered this, thanks for the fix and feedback. Looks good to me.

TGlas:
Thanks for the proposal, I merged it.

@Jartreg On private fields: it seems that we now have two language-level mechanisms for solving a single problem, with differences in the implementation. That's a bad situation for a language (C++ knows what I mean), but it is a good thing that Javascript picks up more and more features from Typescript. Maybe we will get rid of Typescript in a decade or so if JS keeps evolving. TS will not really die, but become assimilated, and then it will have completed its heroic job of making JS great again (maybe it should add operator overloading next :D). Although I personally prefer the keyword syntax, I believe that the newer JS syntax for private fields will prevail in the long run, and now that JS has its own solution it seems very unlikely that the `private` keyword will become legal ECMAScript. For the time being, I don't see a need to change existing code, but it seems more future-proof to use the new feature.

PR #143, opened 2025-05-22
Cleanup of program state and editor handling

Jartreg:
This refactor mostly changes two aspects: how program state is handled and where the logic concerning editors is located.

(Since there are a lot of changes, I hope it's clear enough what is changed in each commit. I tried to group related changes to make review easier…)

# Program State

There are currently two functions (`updateStatus` and `updateControls`) that are supposed to update the UI to reflect the current state of the program. They are called in various locations for non-obvious reasons, which makes it very hard to keep track of when it is necessary to call them. (e.g. when opening a file, `cmd_load` calls `updateControls`. Why? The program state didn't change‽ The only possible reason I could find: `updateControls` calls `updateStatus`, which makes all editors readonly if a program is running. So this could be a roundabout way to make the new editor readonly…)

I refactored this to be (in my opinion) simpler and more bug-resistant: there's a new function [`updateProgramState`](https://github.com/Jartreg/tscript/blob/53694a865e749b60b4af4940790ec84b09a495c9/src/ide/elements/index.ts#L316-L367) in `ide/elements/index.ts` that takes care of everything. My goal was to make the data flow more clear and make it obvious where it needs to be called.

```mermaid
flowchart LR
    A[getProgramStatus]
    interpreterSession@{ shape: lean-r } --> A
    hasErrorMessages@{ shape: lean-r } --> A

    B[updateProgramState]
    S@{ shape: sm-circ } -->|interpreterChanged| B
    A -->|current| B
    currentProgramState@{ shape: lean-r } -->|previous| B
```

- `getProgramStatus` determines the current status by looking at `interpreterSession` and `hasErrorMessage`
- `updateProgramState` uses only the current/previous status to determine what it needs to do
  - Exception: stepping info is always updated when visible
  - Exception: there is a flag to determine when a new interpreter was created
- `updateProgramState` is only called from locations where it's obvious that some of its inputs have changed

# EditorController

`Document.properties` was an untyped object that had properties and methods related to editors. Instead of storing this _in_ the editor view, I moved things into a new class that _contains_ the editor view:

- Everything from `Document.properties` now lives in `EditorController`
- `EditorController` now handles things like saving and breakpoints, which simplified code elsewhere
- Editor events are now typed

# Drive-by Bugfixes/Features

In some cases, the IDE previously didn't clear the stepping information when starting a new program. This is fixed with `updateProgramState`.

<details>
<summary>Demo</summary>

The stack and program tree of will-error are still shown when br-test is run:

<table><tr><th>Before</th><th>After</th></tr>
<tr><td>

https://github.com/user-attachments/assets/bda501cc-0e24-48ee-a55f-a7e55f25bd0d

</td><td>

https://github.com/user-attachments/assets/95d9939e-78d6-40ab-9160-547d78d925a0

</td></tr></table>

</details>

There was an off-by-one error when the interpreter tried to move breakpoints to valid locations: it never found breakpoints on the last line. I fixed this while adding `EditorController`.

<details>
<summary>Demo</summary>

On the left, the program runs to completion despite the breakpoint before the last statement.

<table><tr><th>Before</th><th>After</th></tr>
<tr><td>

https://github.com/user-attachments/assets/ab160a35-4d2d-426b-a659-b508c32d7e96

</td><td>

https://github.com/user-attachments/assets/6fe94ad4-7536-4058-bfa7-6393aada5734

</td></tr></table>

</details>

I found it a little annoying that the IDE always asks for confirmation when closing the tab, even when all documents are saved. Now the confirmation only shows if any EditorController has unsaved changes ([here](https://github.com/Jartreg/tscript/blob/53694a865e749b60b4af4940790ec84b09a495c9/src/ide/index.ts#L80-L83)).

<details>
<summary>Demo</summary>

<table><tr><th>Before</th><th>After</th></tr>
<tr><td>

https://github.com/user-attachments/assets/936b32f9-2a68-43a1-aef9-debf4ea0a704

</td><td>

https://github.com/user-attachments/assets/bb6e398c-92f1-4c60-9f14-fbc050e57963

</td></tr></table>

</details>

When importing a file or opening the IDE with `?load=…`, existing files with the same name were silently overwritten. This is [fixed](https://github.com/TGlas/tscript/pull/143/commits/af34592485bcda5a0896951d3f5ab985e37fc58d) by the new `importData` function.


TGlas:
Thanks once more for your contribution! Again, that makes the structure much cleaner, and the code base easier to work with. Highly appreciated!

PR #150, opened 2025-07-25
Add Git integration

PplaysMine:
Add the Git integration as part of the SS25 study project.


PplaysMine:
The projects-fs test file has to be excluded from `npm run test` due to missing browser APIs,
therefore the file is named `projects-fs-test.ts` instead of `projects-fs.test.ts`

TGlas:
> Currently, our backend server works with http. The goal is to eventually SSL encrypt the server URL so that all traffic is handled via https. To do this, we need to set up automatic certificate renewal, which would go beyond the scope of our study project. We will however update the server in due time so that all traffic is handled through https.

The togla.ini.rub.de server is running let's encrypt anyway (it runs various services), so that would not be a big deal. There is no need to take care of the renewal on your side. All I need is additional fields in the config for entering the paths to the certificates.

TGlas:
> I have updated the server to run on https if key & cert locations are provided in the server config. Only the URLs in github_creds.ts would need to be updated to https://...

Thanks! The server should now be running with SSL enabled.

PR #152, opened 2025-07-26
Two fixes concerning namespaces

ole0811sch:
Fix two bugs with namespaces and fix the lattice craft test.

1. For asserting that the access to a member variable (of a namespace or a static variable of a class) can only be made after that variable is declared, `op.where.pos` was used instead of `op.where.totalpos`, which  breaks when using includes. For example, if these are two files, File1 can't be executed (even though it can when removing the semicolons from line 1 of File2):

File1:
```
include "File2";
print(ns.v);
```
File2:
```
namespace ns {;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
	var v;
}
```

2.  In parse_expression.ts, `ex` was still used in some places in anonymous functions (`passResolve` and `passResolveBack`) even though due to how scoping works, the value of `ex` will always be the last member access expression (from the last iteration of the inner while loop). This led in particular to the following program not being accepted:
```
namespace A {
	namespace B {
		function func(){}
	}
}

A.B.func();
```

The second bug also broke lattice craft. However, the test that checks that lattice craft can be parsed was also broken. The last commit fixes the test.

### DIFF
--- a/src/lang/parser/parse_expression.ts
+++ b/src/lang/parser/parse_expression.ts
@@ -942,7 +942,8 @@ export function parse_expression(
 									if (
 										reference.petype === "variable" &&
 										reference.where.hasOwnProperty("pos") &&
-										op.where.pos < reference.where.pos
+										op.where.totalpos <
+											reference.where.totalpos
 									) {
 										state.error("/name/ne-5", [
 											"expression",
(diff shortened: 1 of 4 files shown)


TGlas:
Good catch! Thanks for the fix.

